PR #57, opened 2026-03-13
feat: Operation-based tools with queue and formatting facade

aaronsb:
## Summary

Refactor from 12 individual MCP tools to 5 operation-based tools, borrowing patterns from the jira-cloud MCP server.

- **5 tools, 17 operations** — `manage_accounts` (3 ops), `manage_email` (5), `manage_calendar` (5), `manage_drive` (4), `queue_operations`
- **Queue handler** — chain operations sequentially with `$N.field` result references and per-operation error strategies (bail/continue)
- **Formatting facade** — response shaping via `markdown.ts`, contextual next-steps guidance via `next-steps.ts`
- **Per-domain handlers** — accounts, email, calendar, drive each in their own file, all under 100 lines
- **No file over 130 lines** — clean decomposition

### Tool Surface
```
manage_accounts  — list | authenticate | remove
manage_email     — search | read | send | reply | triage
manage_calendar  — list | agenda | create | get | delete
manage_drive     — search | upload | get | download
queue_operations — batch with $N.field references
```

## Test plan
- [x] `make typecheck` — clean
- [x] `make build` — clean
- [x] `make test-unit` — 50 tests passing
- [x] MCP smoke test: 5 tools listed with correct operation enums

### DIFF
--- a/src/__tests__/server/queue.test.ts
+++ b/src/__tests__/server/queue.test.ts
@@ -1,0 +1,125 @@
+import { handleQueue } from '../../server/queue.js';
+
+// Simple mock handlers
+const handlers: Record<string, (p: Record<string, unknown>) => Promise<unknown>> = {
+  tool_a: async (params) => ({ id: 'result-a', value: params.input ?? 'default' }),
+  tool_b: async (params) => ({ id: 'result-b', ref: params.ref ?? 'none' }),
+  tool_fail: async () => { throw new Error('intentional failure'); },
+};
+
+describe('handleQueue', () => {
+  it('executes operations sequentially', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_a', args: { input: 'hello' } },
+        { tool: 'tool_b', args: { ref: 'world' } },
+      ],
+    }, handlers) as any;
+
+    expect(result.summary).toEqual({ total: 2, succeeded: 2, failed: 0, skipped: 0 });
+    expect(result.results[0].data).toEqual({ id: 'result-a', value: 'hello' });
+    expect(result.results[1].data).toEqual({ id: 'result-b', ref: 'world' });
+  });
+
+  it('resolves $N.field references', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_a', args: { input: 'first' } },
+        { tool: 'tool_b', args: { ref: '$0.id' } },
+      ],
+    }, handlers) as any;
+
+    expect(result.results[1].data).toEqual({ id: 'result-b', ref: 'result-a' });
+  });
+
+  it('bails on error by default', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_a', args: {} },
+        { tool: 'tool_fail', args: {} },
+        { tool: 'tool_b', args: {} },
+      ],
+    }, handlers) as any;
+
+    expect(result.summary).toEqual({ total: 3, succeeded: 1, failed: 1, skipped: 1 });
+    expect(result.results[1].status).toBe('error');
+    expect(result.results[2].status).toBe('skipped');
+  });
+
+  it('continues on error when onError is continue', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_a', args: {} },
+        { tool: 'tool_fail', args: {}, onError: 'continue' },
+        { tool: 'tool_b', args: {} },
+      ],
+    }, handlers) as any;
+
+    expect(result.summary).toEqual({ total: 3, succeeded: 2, failed: 1, skipped: 0 });
+    expect(result.results[2].status).toBe('success');
+  });
+
+  it('errors on forward references', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_a', args: { input: '$1.id' } },
+        { tool: 'tool_b', args: {} },
+      ],
+    }, handlers) as any;
+
+    expect(result.results[0].status).toBe('error');
+    expect(result.results[0].error).toContain("hasn't run yet");
+  });
+
+  it('errors on references to failed operations', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_fail', args: {}, onError: 'continue' },
+        { tool: 'tool_b', args: { ref: '$0.id' } },
+      ],
+    }, handlers) as any;
+
+    expect(result.results[1].status).toBe('error');
+    expect(result.results[1].error).toContain('error');
+  });
+
+  it('errors on missing field in reference', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'tool_a', args: {} },
+        { tool: 'tool_b', args: { ref: '$0.nonexistent' }, onError: 'continue' },
+      ],
+    }, handlers) as any;
+
+    expect(result.results[1].status).toBe('error');
+    expect(result.results[1].error).toContain('not found');
+  });
+
+  it('errors on unknown tool', async () => {
+    const result = await handleQueue({
+      operations: [
+        { tool: 'unknown_tool', args: {} },
+      ],
+    }, handlers) as any;
+
+    expect(result.results[0].status).toBe('error');
+    expect(result.results[0].error).toContain('Unknown tool');
+  });
+
+  it('strips next_steps from results', async () => {
+    const handlersWithSteps: Record<string, (p: Record<string, unknown>) => Promise<unknown>> = {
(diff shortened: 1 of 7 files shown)


aaronsb:
## Code Review: Operation-Based Tools Refactor

**What this changes**: Consolidates 12 MCP tools into 5 operation-based tools (`manage_accounts`, `manage_email`, `manage_calendar`, `manage_drive`, `queue_operations`), adds sequential queue execution with `$N.field` result references, and attaches response formatting and next-steps guidance to every response.

**Scope**: ~870 lines added, ~470 removed across 12 files. Well-decomposed — no file exceeds 160 lines.

---

### 1. Operation Validation — Required Field Checks

**Assessment: Partially covered, with a gap.**

The handlers do validate operation-specific required fields (e.g., `calendar.ts:39` checks `summary`, `start`, `end` for create; `email.ts:30` checks `to`, `subject`, `body` for send). This is good.

**Issue: `email` not validated as present on domain handlers**

**Location**: `src/server/handlers/email.ts:6`, `calendar.ts:6`, `drive.ts:6`

All three domain handlers do `const email = params.email as string` without checking it exists. The schema marks `email` as required, and MCP SDK *should* enforce that, but defensive validation at the handler level would prevent silent `undefined` propagation if the SDK passes through bad input or if the handler is called via `queue_operations` (which bypasses schema validation entirely).

**Suggestion**: Add an early guard in each domain handler, or extract a shared helper:
```typescript
function requireEmail(params: Record<string, unknown>): string {
  const email = params.email;
  if (typeof email !== 'string' || !email) throw new Error('email is required');
  return email;
}
```

This is especially important given that `queue_operations` calls these handlers directly through `domainHandlers[op.tool](resolvedArgs)` — there is no schema validation on the inner operations.

**Issue: `manage_accounts` with `operation: 'remove'` and no email**

**Location**: `src/server/handlers/accounts.ts:40`

The `remove` operation checks for `email` at runtime (good), but the tool schema in `tools.ts` does not mark `email` as conditionally required for `remove`. This means the MCP client gets no upfront signal that email is needed. The runtime check catches it, but consider adding a note in the `description` field for the email property: "Required for remove".

Wait — the description already says "Required for remove". Good, this is covered descriptively. The runtime check is the real safety net.

---

### 2. Queue Handler Security — `$N.field` References

**Location**: `src/server/queue.ts`

**Issue (Medium): Reference resolution is limited to single-depth and `\w+` field names, which is safe — but the implicit `[0]` array indexing is a hidden behavior.**

`extractField` (lines 107-130) silently returns the first element's `id` when the field is `id` and the result contains `emails`, `events`, or `files` arrays. This means `$0.id` can mean "the id field on the result object" OR "the id of the first item in the result array" depending on the shape of data. This is convenient but:

1. **Undocumented**: The caller has no way to know which behavior will trigger.
2. **Fragile**: If the result has both a top-level `id` AND an `emails` array, the top-level `id` wins (due to the `if (field in obj)` check on line 113). This inconsistency could cause subtle bugs.
3. **Only supports `id`**: What if someone wants `$0.name` from the first file? That falls through to `undefined`.

**Suggestion**: Either:
- (a) Make the array-first-element behavior explicit with syntax like `$0.files[0].id`, or
- (b) Document the current magic behavior clearly in the tool description and add support for more fields beyond just `id`, or
- (c) Remove the magic entirely and require the AI to use explicit structure — the AI is the consumer and can parse results itself.

Option (c) is simplest and most predictable. The magic indexing tries to help but creates an implicit contract that is hard to reason about.

**Issue (Low): No depth limit on nested field access**

The current regex `\$(\d+)\.(\w+)` only captures one level of field access, which is fine. But `resolveRef` uses `String(extracted)` to coerce the result — if `extracted` is an object, this produces `[object Object]`. Consider throwing if the resolved value is not a primitive.

**Security: No injection risk found.** The `$N.field` pattern uses parseInt + array index lookup, field names are restricted to `\w+`, and values are coerced to strings. There is no eval, no dynamic property path traversal beyond one level, and no way to reference fields outside the results array. The `index >= results.length` check prevents forward references. This is solid.

---

### 3. Formatting Quality — Token Efficiency

**Location**: `src/server/formatting/markdown.ts`, `src/server/formatting/next-steps.ts`

**Assessment: Good overall, with one concern.**

The formatters strip raw API verbosity down to essential fields (e.g., email list drops payload details, keeps id/threadId/snippet). This is the right approach for AI token budgets.

**Issue: `next_steps` adds ~200-400 tokens per response**

Every single response includes a `next_steps` array with full example objects. For a queue of 10 operations, that is 10x next_steps payloads in the combined response. The queue use case is precisely where the consumer already knows what to do next (they planned the whole sequence), so next_steps adds noise.

**Suggestion**: Skip `nextSteps()` injection when the handler is called from within `queue_operations`. One approach: add an optional `{ fromQueue?: boolean }` context parameter to each handler, and omit next_steps when true. Or strip next_steps from individual results in `handleQueue` before returning.

**Good pattern**: The `next_steps` structure with `tool` + `example` is well-designed for AI consumption — it provides actionable tool calls, not prose. The placeholder replacement in `replacePlaceholders` is a nice touch for contextualizing the suggestions.

---

### 4. Code Quality

**Assessment: Clean decomposition. No file size concerns.**

All files are well under 200 lines. The handler-per-domain pattern (`handlers/email.ts`, etc.) follows Single Responsibility. The `clamp()` utility is duplicated in three files (`email.ts`, `calendar.ts`, `drive.ts`) — minor, but worth extracting to a shared utils module if more handlers are added.

**Issue: `toGwsArgs` logic moved from declarative (tools.ts) to imperative (handlers/*.ts)**

The old `tools.ts` had `toGwsArgs` functions co-located with each tool definition — a single place to see "this tool maps to these CLI args". The new code spreads this mapping across four handler files. This is a reasonable trade-off (handlers own their full lifecycle), but it means understanding the gws CLI mapping now requires reading four files instead of one.

Not a blocker — just noting the trade-off. The handlers are small enough that this is manageable.

**Issue: `server.ts` lost the "unknown tool" error response**

**Location**: `src/server/server.ts:34`

The old code had an explicit check: `if (!tool) return { content: [...], isError: true }`. The new code delegates to `handleToolCall`, which throws `Unknown tool: ${name}`. This throw is caught by the generic error handler, but it becomes an unhandled exception rather than a clean MCP error response — the generic catch only handles `GwsError`, not plain `Error`.

**Suggestion**: The catch block at `server.ts:45` should handle the general `Error` case too, not just `GwsError`. Currently, a plain `Error` (like "Unknown tool") falls through to the final catch which returns `{ error: String(err) }` — this works but loses structure. Consider:
```typescript
} catch (err) {
  const message = err instanceof Error ? err.message : String(err);
  const detail = err instanceof GwsError
    ? { error: message, exitCode: err.exitCode, reason: err.reason, stderr: err.stderr }
    : { error: message };
  return { content: [{ type: 'text', text: JSON.stringify(detail, null, 2) }], isError: true };
}
```

---

### 5. Test Coverage Gaps

**Assessment: Schema tests are present but behavioral tests are missing.**

The test file validates schema structure (enum values, required fields, additionalProperties). This is useful for catching schema regressions.

**Missing tests:**

| Area | What's untested | Risk |
|------|----------------|------|
| **Queue handler** | `resolveReferences`, `extractField`, bail/continue logic, forward reference rejection | The most complex new code has zero tests |
| **Handler validation** | Per-operation required field checks (e.g., send without `to`) | Runtime validation is the only defense since queue bypasses schema |
| **Formatting** | `formatEmailList`, `formatEventList`, `formatFileList` with various input shapes | Formatters do unsafe casting (`data as Record<string, unknown>`) that could throw on unexpected input |
| **next-steps** | `nextSteps()` with unknown domain/operation, `replacePlaceholders` | Edge cases like missing context keys |

**Priority**: Queue handler tests are the highest priority. The reference resolution logic (`$N.field`) is the most novel and most likely to have edge-case bugs. Suggested test cases:

- `$0.id` resolves from a direct field
- `$0.id` resolves from `files[0].id` fallback
- `$2.field` throws when only 2 results exist (off-by-one)
- `$0.field` throws when operation 0 failed
- Mixed `bail` and `continue` error strategies
- String with multiple references: `"From $0.id to $1.id"`
- Non-string args are passed through unchanged

---

### Summary

| Category | Verdict |
|----------|---------|
| Architecture | Good — clean decomposition, SRP followed |
| Operation validation | Adequate with gaps — queue bypasses schema validation |
| Queue security | Safe — no injection vectors found |
| Token efficiency | Good formatters, but next_steps bloats queue responses |
| Test coverage | Insufficient — queue handler and formatters need tests |

**Recommendation**: Address the queue test coverage and the queue-bypasses-schema-validation issue before merging. The other items are improvements that can follow.

PR #59, opened 2026-03-15
test: Strengthen unit tests and add integration tests

aaronsb:
## Summary

- **55 new unit tests** covering previously untested critical paths: server error mapping, handler dispatch routing, input validation edge cases, credential file permissions (0600/0700), and executor timeout behavior
- **10 integration tests** that hit real Google APIs through gws with real credentials — no mocks, full stack verification
- **Self-documenting Makefile** with `help` as default target

## Test inventory

| Suite | Tests | What it covers |
|-------|-------|---------------|
| Unit (existing) | 114 | Executor, handlers, formatters, queue, schemas |
| Unit (new) | 55 | validate.ts edges, server.ts error mapping, handler dispatch, credential permissions, executor timeout |
| Integration | 10 | Real gws → Google API calls for email, calendar, drive, accounts |
| **Total** | **179** | |

## Usage

```
make                                        # show targets
make test                                   # unit tests (~1s)
make test-integration                       # integration (~5s, uses first account)
make test-integration ACCOUNT=user@example  # explicit account override
make test-all                               # unit + integration
```

## Test plan

- [x] `make test` — 169 unit tests pass
- [x] `make test-integration` — 10 integration tests pass with default account
- [x] `make test-integration ACCOUNT=aaronsb@gmail.com` — explicit override works
- [x] `make test-all` — both suites pass
- [x] Integration tests auto-skip when no credentials available

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ test-unit: ## Run unit tests (mocked, fast, no network)
 	npx jest --config jest.config.cjs --testPathPattern='src/__tests__/(executor|accounts|server)' --runInBand
 
 test-integration: ## Run integration tests (ACCOUNT=email optional)
-	TEST_ACCOUNT=$(ACCOUNT) npx jest --config jest.config.cjs --testPathPattern='src/__tests__/integration' --runInBand
+	$(if $(ACCOUNT),TEST_ACCOUNT=$(ACCOUNT)) npx jest --config jest.config.cjs --testPathPattern='src/__tests__/integration' --runInBand
 
 typecheck: ## Type-check without emitting
 	npx tsc --noEmit --skipLibCheck
--- a/src/__tests__/accounts/credentials.test.ts
+++ b/src/__tests__/accounts/credentials.test.ts
@@ -38,10 +38,20 @@ describe('credentials', () => {
         'utf-8',
       );
     });
+
+    it('throws when credential file does not exist', async () => {
+      mockFs.readFile.mockRejectedValue(new Error('ENOENT'));
+      await expect(credentials.readCredential('user@example.com')).rejects.toThrow('ENOENT');
+    });
+
+    it('throws on malformed JSON', async () => {
+      mockFs.readFile.mockResolvedValue('not-json{{{');
+      await expect(credentials.readCredential('user@example.com')).rejects.toThrow();
+    });
   });
 
   describe('removeCredential', () => {
-    it('deletes credential file', async () => {
+    it('deletes credential file at the correct path', async () => {
       mockFs.unlink.mockResolvedValue(undefined);
       await credentials.removeCredential('user@example.com');
       expect(mockFs.unlink).toHaveBeenCalledWith(credentialPath('user@example.com'));
@@ -59,6 +69,11 @@ describe('credentials', () => {
       err.code = 'EACCES';
       mockFs.unlink.mockRejectedValue(err);
       await expect(credentials.removeCredential('user@example.com')).rejects.toThrow('EACCES');
+    });
+
+    it('rethrows errors without code property', async () => {
+      mockFs.unlink.mockRejectedValue(new Error('unknown fs error'));
+      await expect(credentials.removeCredential('user@example.com')).rejects.toThrow('unknown fs error');
     });
   });
 
--- a/src/__tests__/executor/gws.test.ts
+++ b/src/__tests__/executor/gws.test.ts
@@ -126,18 +126,19 @@ describe('execute', () => {
 
   it('rejects with timeout error and sends SIGTERM', async () => {
     jest.useFakeTimers();
-    const proc = createMockProcess();
-    mockSpawn.mockReturnValue(proc as any);
+    try {
+      const proc = createMockProcess();
+      mockSpawn.mockReturnValue(proc as any);
 
-    const promise = execute(['calendar', 'events', 'list'], { timeout: 100 });
+      const promise = execute(['calendar', 'events', 'list'], { timeout: 100 });
 
-    // Advance past the timeout
-    jest.advanceTimersByTime(150);
+      jest.advanceTimersByTime(150);
 
-    await expect(promise).rejects.toThrow('timed out');
-    expect(proc.kill).toHaveBeenCalledWith('SIGTERM');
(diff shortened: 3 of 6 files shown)


aaronsb:
## Code Review: test/strengthen-unit-tests

854 additions across 8 test files + Makefile improvements. Reviewed all new/modified files against the source code they cover.

---

### What this PR does well

The test suite is substantially more robust after this change. Several highlights:

- **`validate.test.ts`** is excellent security-conscious testing. Path traversal checks (`../etc/passwd`), type coercion edge cases, and boundary inputs for email validation are exactly what input validation tests should cover.
- **`server.test.ts`** tests the actual MCP protocol wiring — GwsError to structured `isError` response mapping, non-Error thrown values, argument defaulting. These verify real server behavior, not just pass-through.
- **`handler.test.ts`** verifies routing isolation (calling `manage_email` does NOT invoke `handleAccounts`), which is a meaningful contract test.
- **Integration test skip pattern** (`describeIf = account ? describe : describe.skip`) is clean and CI-safe.
- **Makefile** self-documenting `help` target follows the standard pattern nicely.

---

### Issues to consider

#### 1. Bug discovery: `clamp` has surprising behavior with 0 and negatives

**Location**: `src/server/handlers/validate.ts:20` / `src/__tests__/server/handlers/validate.test.ts:93-109`

The test at line 93 documents that `clamp(0, 10, 50)` returns `10` (the default), not `0`. And at line 107, `clamp(-5, 10, 50)` returns `-5` — below the "default" of 10. The test at line 105 shows `clamp(1, 10, 50)` returns `1`, which is also below the default.

The function uses `Number(value) || defaultVal` which means:
- `0` is treated as missing (falsy) — probably a bug if someone explicitly passes `maxResults: 0`
- Values below `defaultVal` pass through unclamped — the parameter name `defaultVal` is actually acting as a **minimum** for falsy values but not for small truthy values
- There is no lower bound enforcement at all

The tests correctly document this behavior, but this looks like a latent bug in the source code rather than intended behavior. The function name `clamp` implies bounding to a range, but it only enforces a ceiling. Consider whether this should be flagged as a known issue or fixed. At minimum, the test comments should explicitly call out "this documents current behavior, which may be surprising."

#### 2. Integration test credential path leaks real filesystem structure

**Location**: `src/__tests__/integration/setup.ts:53`

When `TEST_ACCOUNT` is set but the credential file doesn't exist, the warning message includes the full credential file path:
```
TEST_ACCOUNT=foo@bar.com but no credential file at /home/user/.local/share/google-workspace-mcp/credentials/foo_at_bar_dot_com.json
```

This is a `console.warn` so it only appears in test output, not in production. Low risk, but worth noting — if CI logs are public, this reveals the local filesystem layout. Consider using a relative path or omitting the full path from the warning.

#### 3. `server.test.ts` relies on handler registration order

**Location**: `src/__tests__/server/server.test.ts:72-75`

```typescript
const calls = mockSetRequestHandler.mock.calls;
listToolsHandler = calls[0][1];
callToolHandler = calls[1][1];
```

This assumes `ListToolsRequestSchema` is registered before `CallToolRequestSchema`. It matches the current source, but if someone reorders the `setRequestHandler` calls in `server.ts`, these tests silently test the wrong handlers. More robust approach: find the call by matching the schema argument:

```typescript
listToolsHandler = calls.find(c => c[0] === 'ListToolsRequestSchema')![1];
callToolHandler = calls.find(c => c[0] === 'CallToolRequestSchema')![1];
```

#### 4. Integration handler tests use broad `as any` casting

**Location**: `src/__tests__/integration/handlers.test.ts` — lines 16, 31, 44, etc.

Every handler result is cast to `any` before assertions. This means the tests won't catch if handler return types change. For example, if `handleEmail` starts returning `{ messages: [] }` instead of `{ emails: [] }`, the test still compiles and only fails at runtime.

Since these are integration tests validating shape contracts, this is somewhat expected — but consider defining lightweight `Expected*` interfaces for the return shapes. This would make the tests serve double duty as documentation of the handler contract.

#### 5. Missing coverage: `readCredential` and `removeCredential`

**Location**: `src/accounts/credentials.ts:38-50`

The `exportAndSaveCredential` tests are thorough (happy path, wrong type, null data, empty object). But `readCredential` and `removeCredential` remain untested. `removeCredential` in particular has an ENOENT-swallowing catch that would be good to verify — both the "file exists" and "file missing" paths, plus re-throwing non-ENOENT errors.

#### 6. Makefile: `TEST_ACCOUNT=$(ACCOUNT)` passes empty string when unset

**Location**: `Makefile:16`

```makefile
TEST_ACCOUNT=$(ACCOUNT) npx jest ...
```

When `ACCOUNT` is not provided, this sets `TEST_ACCOUNT=` (empty string) in the environment, not unset. The integration setup handles this correctly because `process.env.TEST_ACCOUNT` will be `""` which is falsy — so it works. But it is technically different from the env var being absent. Consider:

```makefile
test-integration: ## Run integration tests (ACCOUNT=email optional)
ifdef ACCOUNT
	TEST_ACCOUNT=$(ACCOUNT) npx jest ...
else
	npx jest ...
endif
```

This is minor — current behavior works — but it's cleaner.

#### 7. `executor.test.ts` timeout test: timer cleanup on failure path

**Location**: `src/__tests__/executor/gws.test.ts:123-137`

```typescript
it('rejects with timeout error and sends SIGTERM', async () => {
    jest.useFakeTimers();
    // ...
    jest.useRealTimers();
});
```

If the test assertion fails (throws), `jest.useRealTimers()` never runs, potentially leaving fake timers active for subsequent tests. Use a `try/finally` or move the timer setup/teardown to `beforeEach`/`afterEach` within a nested `describe`.

---

### Assessment

**Solid PR.** The tests verify meaningful behavior — error mapping, routing isolation, security validation, file permission enforcement, timeout handling. These are not hollow "assert truthy" tests. The integration test skip mechanism is well-designed for CI without credentials.

The `clamp` behavior (item 1) is the only thing I'd call a potential bug in production code surfaced by these tests. Everything else is minor improvement opportunities. No security issues, no credential leakage in test fixtures.

**Recommendation**: Merge-ready. Consider addressing items 1 and 7 as follow-ups.

PR #60, opened 2026-03-15
feat: Markdown formatting facade for LLM agent consumption

aaronsb:
## Summary

- **Replaces JSON responses with token-efficient markdown** — lists use pipe-delimited compact rows, detail views use natural prose agents can relay to users
- **Dual-return pattern** (`{ text, refs }`) — handlers return markdown for the agent plus structured refs for queue `$N.field` chaining
- **Next-steps as markdown footer** — appended to every response instead of a separate JSON key
- **Queue renders summary with status icons** — `✓`, `✗`, `○` per operation with consolidated next-steps

Patterns borrowed from jira-cloud MCP per project conventions (operation-based tools, formatting facade, next-steps guidance).

## Before/After

**Email list (before):**
```json
{ "emails": [{ "id": "msg-1", "from": "alice@test.com", "subject": "Hello" }], "count": 1 }
```

**Email list (after):**
```markdown
## Messages (3)

msg-1 | alice@test.com | Hello | Mar 14
msg-2 | bob@test.com | Meeting notes | Mar 13

---
**Next steps:**
- Read a specific email: `manage_email` — `{ "operation": "read", ... }`
```

## Test plan

- [x] `make test` — 176 unit tests pass
- [x] `make test-integration` — 10 integration tests pass against real Google APIs
- [x] `make test-all` — 186 total, all green
- [x] Queue `$N.field` resolution works via refs (not fragile regex on markdown)

### DIFF
--- a/src/server/formatting/markdown.ts
+++ b/src/server/formatting/markdown.ts
@@ -9,14 +9,15 @@
  *   structured values queue $N.field resolution needs
  */
 
-export interface FormattedResponse {
+/** Shared response shape — markdown text for agents, structured refs for queue $N.field. */
+export interface HandlerResponse {
   text: string;
   refs: Record<string, unknown>;
 }
 
 // --- Email formatting ---
 
-export function formatEmailList(data: unknown): FormattedResponse {
+export function formatEmailList(data: unknown): HandlerResponse {
   const raw = data as Record<string, unknown>;
   const messages = (raw?.messages ?? raw?.items ?? []) as Record<string, unknown>[];
 
@@ -45,7 +46,7 @@ export function formatEmailList(data: unknown): FormattedResponse {
   };
 }
 
-export function formatEmailDetail(data: unknown): FormattedResponse {
+export function formatEmailDetail(data: unknown): HandlerResponse {
   const msg = data as Record<string, unknown>;
   const payload = msg.payload as Record<string, unknown> | undefined;
   const headers = (payload?.headers ?? []) as Array<{ name: string; value: string }>;
@@ -90,7 +91,7 @@ export function formatEmailDetail(data: unknown): FormattedResponse {
 
 // --- Calendar formatting ---
 
-export function formatEventList(data: unknown): FormattedResponse {
+export function formatEventList(data: unknown): HandlerResponse {
   const raw = data as Record<string, unknown>;
   const items = (raw?.items ?? []) as Record<string, unknown>[];
 
@@ -123,7 +124,7 @@ export function formatEventList(data: unknown): FormattedResponse {
   };
 }
 
-export function formatEventDetail(data: unknown): FormattedResponse {
+export function formatEventDetail(data: unknown): HandlerResponse {
   const event = data as Record<string, unknown>;
   const attendees = (event.attendees ?? []) as Array<Record<string, unknown>>;
   const id = String(event.id ?? '');
@@ -179,7 +180,7 @@ export function formatEventDetail(data: unknown): FormattedResponse {
 
 // --- Drive formatting ---
 
-export function formatFileList(data: unknown): FormattedResponse {
+export function formatFileList(data: unknown): HandlerResponse {
   const raw = data as Record<string, unknown>;
   const items = (raw?.files ?? []) as Record<string, unknown>[];
 
@@ -208,7 +209,7 @@ export function formatFileList(data: unknown): FormattedResponse {
   };
 }
 
-export function formatFileDetail(data: unknown): FormattedResponse {
+export function formatFileDetail(data: unknown): HandlerResponse {
   const file = data as Record<string, unknown>;
   const id = String(file.id ?? '');
   const name = String(file.name ?? '');
--- a/src/server/handler.ts
+++ b/src/server/handler.ts
@@ -4,13 +4,8 @@ import { handleCalendar } from './handlers/calendar.js';
 import { handleDrive } from './handlers/drive.js';
 import { handleQueue } from './queue.js';
 
-/**
- * Handlers return text for the agent plus refs for queue $N.field resolution.
- */
-export interface HandlerResponse {
-  text: string;
-  refs: Record<string, unknown>;
-}
+export type { HandlerResponse } from './formatting/markdown.js';
+import type { HandlerResponse } from './formatting/markdown.js';
 
 type ToolHandler = (params: Record<string, unknown>) => Promise<HandlerResponse>;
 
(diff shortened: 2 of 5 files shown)


aaronsb:
## Code Review: Formatting Facade PR

**What this changes:** Replaces JSON handler responses with token-efficient markdown, introducing a dual-return `{ text, refs }` contract across all handlers. Queue `$N.field` resolution now reads from structured `refs` instead of parsing JSON text. Next-steps guidance is appended as a markdown footer.

---

### What looks solid

**Clean separation of concerns.** The formatting layer (`markdown.ts`) is a pure data-in/markdown-out facade with no side effects, no imports from handlers, and no awareness of MCP protocol. Handlers orchestrate (call executor, pick formatter, append next-steps), formatters shape output, and queue composes results. This is textbook Single Responsibility.

**The `{ text, refs }` contract is well-designed.** Splitting agent-facing prose from machine-readable fields for queue chaining is the right move. It avoids the fragility of regex-parsing markdown to extract IDs, and it means formatters can change their text output without breaking queue resolution.

**Queue `$N.field` resolution is robust.** Forward-reference detection, failed-operation guarding, missing-field errors -- all the edge cases are covered and tested. The `stripNextSteps` / `extractNextSteps` approach to consolidate next-steps from the last successful operation is a nice touch.

**File sizes are well-controlled.** `markdown.ts` at 321 lines, `queue.ts` at 170, `next-steps.ts` at 126. Handlers are 72-93 lines each. Nothing approaching the 300+ warning threshold.

---

### Issues to address

#### 1. Duplicate `{ text, refs }` interface definitions

**Location:** `src/server/formatting/markdown.ts:12-15` and `src/server/handler.ts:10-13`

**Problem:** `FormattedResponse` and `HandlerResponse` are structurally identical interfaces (`{ text: string; refs: Record<string, unknown> }`) defined in two different files. They aren't linked -- `FormattedResponse` is exported from `markdown.ts` but never imported by handlers, which use `HandlerResponse` instead.

**Why it matters:** If one evolves (say `refs` gets a stricter type), the other stays stale. A reader would reasonably wonder if they're interchangeable or subtly different.

**Suggestion:** Export a single canonical interface from `handler.ts` (or a shared types file) and import it in `markdown.ts`. Or have `FormattedResponse` extend/alias `HandlerResponse`. The formatters and handlers return the same shape -- make the type system reflect that.

#### 2. Queue `refs` loses per-operation references

**Location:** `src/server/queue.ts:110-117`

**Problem:** The queue's returned `refs` only contains `{ total, succeeded, failed, skipped }`. The per-operation `refs` (stored in each `OperationResult`) are not exposed. This means a caller can't reference queue results via `$N.field` in a *subsequent* queue call, and the queue's own refs lack the richness of individual operation results.

**Why it matters:** The plan document (workstream 2) mentions queue enhancement as a follow-up, so this may be intentional scoping. But as-is, if an agent inspects `result.refs` after a queue call, it only sees counters -- not the `messageId`, `fileId`, etc. from the operations that ran. Worth a comment in the code if this is a deliberate "good enough for now" choice.

**Suggestion:** Either merge per-operation refs (namespaced as `op0_messageId`, `op1_fileId` or similar), or at minimum add a code comment noting this is deferred to queue-enhancement.

#### 3. `agenda` handler has a weak refs contract

**Location:** `src/server/handlers/calendar.ts:34-41`

**Problem:** The `agenda` operation does `refs: typeof data === 'object' && data !== null ? data as Record<string, unknown> : {}`. This blindly casts the entire `+agenda` response as refs. Unlike every other handler path, there's no curated set of fields -- whatever the CLI returns becomes the ref bag.

**Why it matters:** This is a leaky abstraction. If `gws calendar +agenda` changes its output shape, queue `$N.field` references against an agenda operation would silently break or resolve to unexpected values. Every other handler path carefully constructs `refs` with named fields.

**Suggestion:** Either format the agenda response through `formatEventList` (if it returns the same shape), or explicitly pick the fields you want in refs. If the agenda format is genuinely opaque, return `refs: {}` and add a comment explaining why.

#### 4. `accounts` handler builds markdown inline instead of using a formatter

**Location:** `src/server/handlers/accounts.ts:17-30`

**Problem:** The accounts `list` operation constructs its markdown directly in the handler (`lines.map(...)`, `## Accounts (${...})`), while email/calendar/drive handlers delegate to `markdown.ts` formatters. The accounts handler also uses `any` casts (line 17, 27).

**Why it matters:** This breaks the pattern established by the other three domains. If you later want to adjust the accounts list format (e.g., add a badge, change marker style), you have to edit handler logic rather than a formatter function. The `any` casts also bypass type safety.

**Suggestion:** Add `formatAccountList` to `markdown.ts` (or a separate `accounts` formatter) and delegate from the handler. Define a minimal type for account objects to replace `any`.

---

### Minor suggestions (non-blocking)

**`formatEmailList` fallback to `items` array** (`markdown.ts:21`): The `raw?.items` fallback is interesting -- is this for the triage response shape vs. the messages.list shape? A brief comment explaining when `items` vs `messages` occurs would help future readers.

**Truncation length constants** (`markdown.ts:29, 103, 192`): The magic numbers 30, 50, 40 for truncation lengths are scattered. Consider extracting them as named constants at the top of the file (e.g., `const MAX_FROM_LENGTH = 30`). Not urgent, but helps when tuning output width.

**Next-steps placeholder detection** (`next-steps.ts:116-119`): The `replacePlaceholders` function does shallow replacement only -- it won't recurse into nested objects in `example`. Currently all examples are flat, so this is fine, but if someone adds a nested example object, the placeholders inside it would be silently ignored. A comment noting the single-depth assumption would be helpful.

**Test verification depth for `$N.field` resolution** (`queue.test.ts:34-45`): The test for `$N.field` resolution verifies that `tool_b` succeeded but doesn't directly assert what value was resolved. You could have the mock echo back the resolved value (e.g., check `result.text` contains `ref=result-a`) to prove the value actually flowed through.

---

### Assessment

This is a well-structured PR that delivers clear value -- the markdown output is genuinely more useful for LLM agents than raw JSON. The formatting layer is properly separated, file sizes are controlled, and the dual-return pattern is consistently applied across 4 domains. The queue integration is thoughtfully done.

The main structural issue is the duplicate interface definitions (#1) and the accounts handler not following the formatter pattern (#4). Issue #2 (queue refs) is fine if deliberately scoped -- just document it. Issue #3 (agenda refs) should be tightened before this ships.

None of these are merge-blockers on their own, but #1 and #3 together represent a pattern where the type contract is looser than it should be for a system that chains values through `$N.field` resolution.

PR #62, opened 2026-03-18
feat: ADR-200 auth lifecycle — status, refresh, scopes operations

aaronsb:
## Summary

Implements ADR-200: expand `manage_accounts` from 3 operations (list, authenticate, remove) to 6, completing the auth lifecycle so users never need to touch `gws` CLI directly.

- **`status`** — check token validity, authorized scopes, and refresh token presence via `gws auth status`
- **`refresh`** — re-export credentials from gws encrypted store when our plaintext copy is stale
- **`scopes`** — re-authenticate with different service selection via `gws auth login -s <services>`
- **Auth error guidance** — GwsError responses with exit code 2 now append remediation next-steps (re-authenticate, check status)

ADR-200 status updated from Draft → Accepted.

## manage_accounts operations (before → after)

| Before | After |
|--------|-------|
| list | list |
| authenticate | authenticate |
| remove | remove |
| — | **status** |
| — | **refresh** |
| — | **scopes** |

## Test plan

- [x] `make test` — 194 unit tests pass (16 suites, +1 new accounts handler suite)
- [x] `make test-all` — 204 total (unit + integration)
- [x] New accounts handler tests cover status (valid/invalid), refresh, scopes, and validation
- [x] Server test verifies auth error guidance appended to GwsError responses

### DIFF
--- a/docs/architecture/auth/ADR-200-unified-account-lifecycle-management-tool.md
+++ b/docs/architecture/auth/ADR-200-unified-account-lifecycle-management-tool.md
@@ -109,6 +109,14 @@ Auth failures should return specific remediation:
 - Token validity checks add an API call per `status` invocation
 - The `setup` operation only works when `gcloud` CLI is installed
 
+## Deferred
+
+The following items from the original design are deferred to future work:
+
+- **`setup` operation** — requires `gcloud` CLI and is hard to automate for headless/MCP environments. Users should run `gws auth setup` directly for first-time Cloud project creation.
+- **`list` status enrichment** — calling `gws auth status` per account on every list is slow. The current list shows credential file presence; full status is available via the `status` operation.
+- **Client secret as tool parameter** — currently sourced from env vars only. Accepting via tool params would require careful handling to avoid logging secrets in MCP request traces.
+
 ## Alternatives Considered
 
 - **Separate `manage_auth` tool**: Splits account CRUD from auth lifecycle. Rejected because accounts and auth are the same domain — an account without auth is useless.
--- a/src/__tests__/server/handlers/accounts.test.ts
+++ b/src/__tests__/server/handlers/accounts.test.ts
@@ -124,7 +124,7 @@ describe('handleAccounts', () => {
       process.env = { ...originalEnv, GOOGLE_CLIENT_ID: 'test-id', GOOGLE_CLIENT_SECRET: 'test-secret' };
     });
 
-    afterAll(() => {
+    afterEach(() => {
       process.env = originalEnv;
     });
 
--- a/src/accounts/auth.ts
+++ b/src/accounts/auth.ts
@@ -1,7 +1,6 @@
 import { spawn, execFile } from 'node:child_process';
 import { platform } from 'node:os';
 import { execute } from '../executor/gws.js';
-import { credentialPath } from '../executor/paths.js';
 import { exportAndSaveCredential } from './credentials.js';
 
 export interface AuthResult {
@@ -31,10 +30,27 @@ export async function checkAccountStatus(email: string): Promise<AccountStatus> 
   };
 }
 
+export async function authenticateAccount(
+  clientId: string,
+  clientSecret: string,
+): Promise<AuthResult> {
+  return runAuthLogin(clientId, clientSecret, ['auth', 'login']);
+}
+
 export async function reauthWithServices(
   clientId: string,
   clientSecret: string,
   services: string,
+): Promise<AuthResult> {
+  return runAuthLogin(clientId, clientSecret, ['auth', 'login', '-s', services]);
+}
+
+// --- Internal ---
+
+function runAuthLogin(
+  clientId: string,
+  clientSecret: string,
+  args: string[],
 ): Promise<AuthResult> {
   return new Promise((resolve, reject) => {
     const env: Record<string, string> = {
@@ -43,7 +59,7 @@ export async function reauthWithServices(
       GOOGLE_WORKSPACE_CLI_CLIENT_SECRET: clientSecret,
     };
 
-    const proc = spawn('gws', ['auth', 'login', '-s', services], {
+    const proc = spawn('gws', args, {
       env,
       stdio: ['ignore', 'pipe', 'pipe'],
     });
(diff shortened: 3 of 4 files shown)


aaronsb:
## Code Review: ADR-200 Auth Lifecycle

**What this changes**: Expands `manage_accounts` from 3 to 6 operations (adding `status`, `refresh`, `scopes`) and appends auth-aware remediation guidance to `GwsError` responses with exit code 2.

---

### 1. Security

**Credential handling looks solid.** Client secrets are read from `process.env` and passed through spawned process environment — never logged, never serialized into responses. The `execFile` usage for `openBrowser` (auth.ts:92) avoids shell injection on the URL. No secrets leak into the `refs` objects returned to clients.

**One concern — `process.env` spread:**

`src/accounts/auth.ts:41` and `:102` both do `...process.env as Record<string, string>`. This passes the *entire* parent environment to the spawned `gws` process, which is intentional (gws needs PATH, HOME, etc.) but worth noting: if the parent process has sensitive env vars unrelated to gws, they propagate. This is an existing pattern pre-PR, not a regression — just flagging for awareness.

**Unused import:**

`src/accounts/auth.ts:4` — `credentialPath` is imported from `../executor/paths.js` but never called anywhere in the file. It was likely added in anticipation of use in `reauthWithServices` but ended up unnecessary since `exportAndSaveCredential` handles the path internally. Should be removed to avoid confusion.

---

### 2. Error Handling

**`checkAccountStatus` has no error handling for non-existent accounts.**

`src/accounts/auth.ts:23` — If `execute(['auth', 'status'], { account: email })` throws a `GwsError` (e.g., the account doesn't exist in gws's store, or the credential file is missing), that exception propagates raw through `handleAccounts` and hits the generic catch in `server.ts`. The user gets a JSON error blob instead of a helpful "account not found — try `list` first" message.

Suggestion: Either catch `GwsError` in `checkAccountStatus` and return a degraded `AccountStatus` (tokenValid: false, empty scopes), or catch it in the `status` case of `handleAccounts` and throw a more specific error. The former is probably better — a "status" check that errors on invalid accounts is less useful than one that reports "invalid."

**`result.data` null/undefined access:**

`src/accounts/auth.ts:24` — `result.data as Record<string, unknown>` assumes `result.data` is truthy. If `gws auth status` returns valid JSON with no `data` field (or `data: null`), accessing `.user`, `.token_valid`, etc. on the cast would throw a runtime TypeError. A defensive `const data = (result.data ?? {}) as Record<string, unknown>` would be safer.

**`reauthWithServices` stdout JSON parse failure:**

`src/accounts/auth.ts:72` — If `gws auth login -s` succeeds (exit code 0) but returns non-JSON stdout, the catch at line 80 resolves with a status: 'error'. This is reasonable, but the error message includes the raw parse error (`Unexpected token...`) rather than something actionable. Minor, but a wrapper message like "gws auth login succeeded but returned unexpected output" would help debugging.

---

### 3. Type Safety

**The `as any` casts in tests are acceptable here.**

`src/__tests__/server/handlers/accounts.test.ts:24-25` — The `as any` on mock return values is a pragmatic choice. The `Account` type likely has more fields than the test needs to specify. Since these are test fixtures behind a mocked module boundary, the casts don't mask real type issues. If you wanted to tighten this later, a `Partial<EnrichedAccount>` test factory would work, but it's not urgent.

**The `Record<string, string>` cast on `process.env` is technically incorrect** (`process.env` values can be `undefined`), but this is a pervasive Node.js pattern and `spawn` handles undefined values fine. Not worth changing.

---

### 4. ADR Compliance

ADR-200 specifies 7 operations. The implementation delivers 6. The missing one:

- **`setup`** — "First-time Cloud project creation (interactive)" wrapping `gws auth setup`. The ADR describes it as the first step in the credential lifecycle. It's not in the PR.

This is worth documenting — either the ADR should note `setup` as deferred/future-work, or there should be a follow-up issue. Currently the ADR says "Accepted" with `setup` in the table, which implies it was implemented.

ADR-200 also specifies that `list` should return enriched status per account (`tokenValid`, `scopes`, `lastAuthenticated`). The current `list` implementation returns only `hasCredential` — the status enrichment described in the ADR's "Status enrichment" section isn't implemented. This is likely intentional (calling `gws auth status` per account on every `list` would be slow), but the divergence should be noted in the ADR or as a future optimization.

ADR-200 also specifies that `authenticate` should accept client ID/secret as tool parameters (not just env vars) for setup flows. This isn't implemented either.

**Recommendation**: Add a "Implementation Notes" section to ADR-200 listing what was deferred: `setup` operation, `list` status enrichment, client secret as tool params. This keeps the ADR honest.

---

### 5. Test Quality

**Good coverage of the happy paths and validation.** The test suite covers:
- All 6 operations (list, remove, status, refresh, scopes + unknown)
- Empty state (list with no accounts)
- Required field validation (email, services)
- Both valid and invalid token status

**Missing test scenarios:**

- **`authenticate` operation** — No test case in the new suite. It existed before this PR but the new `accounts.test.ts` file covers everything *except* authenticate. Since this file is positioned as the comprehensive handler test, the gap is visible.

- **`scopes` with missing env vars** — The test sets up `GOOGLE_CLIENT_ID`/`GOOGLE_CLIENT_SECRET` in `beforeEach` but doesn't test the case where they're missing (should throw "environment variables are required"). Same gap exists for `authenticate`.

- **`scopes` failure path** — No test for when `reauthWithServices` returns `{ status: 'error', error: '...' }`. The handler branches on `result.status` (accounts.ts:139-141) but only the success path is tested.

- **`afterAll` vs `afterEach`** — `src/__tests__/server/handlers/accounts.test.ts:128` uses `afterAll` to restore `process.env`, but the mutation happens in `beforeEach`. If a test in the `scopes` describe block fails and another describe block runs after, `process.env` remains mutated. Should be `afterEach` for symmetry and safety.

---

### 6. Code Duplication: `reauthWithServices` vs `authenticateAccount`

These two functions (`src/accounts/auth.ts:34-85` and `:97-163`) are structurally identical. The only difference is `reauthWithServices` passes `-s <services>` to `gws auth login`. Side-by-side:

| Aspect | `authenticateAccount` | `reauthWithServices` |
|--------|----------------------|---------------------|
| Spawn args | `['auth', 'login']` | `['auth', 'login', '-s', services]` |
| Env setup | identical | identical |
| stderr URL capture | identical | identical |
| stdout parse + export | identical | identical |
| Error handling | identical | identical |

**Suggestion**: Extract a shared `runAuthLogin(clientId, clientSecret, extraArgs?: string[]): Promise<AuthResult>` and have both functions call it. This eliminates ~50 lines of duplication and ensures future fixes (e.g., better error messages, timeout handling) apply to both paths.

```typescript
// Sketch:
async function runAuthLogin(clientId: string, clientSecret: string, extraArgs: string[] = []): Promise<AuthResult> {
  // shared spawn + parse + export logic
}

export async function authenticateAccount(clientId: string, clientSecret: string): Promise<AuthResult> {
  return runAuthLogin(clientId, clientSecret);
}

export async function reauthWithServices(clientId: string, clientSecret: string, services: string): Promise<AuthResult> {
  return runAuthLogin(clientId, clientSecret, ['-s', services]);
}
```

---

### 7. Additional Observations

**`scope_error` guidance is defined but never triggered:**

`src/server/formatting/next-steps.ts:44-46` defines a `scope_error` key, but `server.ts` only checks for `GwsExitCode.AuthError` (exit code 2). If gws returns a scope-related error with a different exit code, or if the error reason indicates scope issues, the `scope_error` guidance is dead code. Either wire it up (check `err.reason` for scope-related strings) or remove it to avoid confusion.

**`scopes` operation doesn't use the `email` param for the actual reauth:**

`src/server/handlers/accounts.ts:138` — `reauthWithServices(clientId, clientSecret, services)` doesn't receive `email`. The reauth spawns `gws auth login -s <services>` which starts a fresh browser flow — it doesn't target a specific account. The email is only used for the response text and next-steps context. This means calling `scopes` with `email: "alice@example.com"` could end up re-authing a *different* account (whichever the user logs into in the browser). This is potentially confusing. At minimum, the tool description should clarify this behavior. Ideally, `reauthWithServices` would accept email and verify the resulting account matches.

---

### Assessment

This is solid incremental work that meaningfully extends the auth lifecycle. The code is well-structured, follows existing patterns, and the next-steps guidance is a nice touch for agent-driven flows.

**Should address before merge:**
1. Remove unused `credentialPath` import from auth.ts
2. Fix `afterAll` -> `afterEach` in scopes test block
3. Add a note to ADR-200 about deferred items (setup, list enrichment, client secret params)

**Worth addressing soon (follow-up OK):**
4. Error handling in `checkAccountStatus` for non-existent accounts
5. Extract shared auth login logic to eliminate duplication
6. Email mismatch risk in `scopes` operation
7. Wire up or remove `scope_error` guidance

PR #78, opened 2026-03-27
feat(scratchpad): service-agnostic content authoring buffer (ADR-301)

aaronsb:
## Summary

- Adds `manage_scratchpad` MCP tool — an in-memory, line-addressed content buffer that decouples composition from delivery
- 17 operations: line editing, JSON path editing, attachments, import from 5 GWS sources, send to 8 GWS targets
- Epoch-based GC (100 tool calls), late-bound targets, cross-scratchpad assembly, format-driven validation
- ADR-301 documents the design: late-bound targets, live binding for JSON mode, attachment side-table

**Send targets:** email, email_draft, doc_create, doc_write, workspace, sheet_write, calendar_event, task_create
**Import sources:** doc (markdown/json), email, sheet, drive_file, meet transcript

## Why

Every content operation was previously a direct API call — no retry resilience, no incremental editing, no content reuse across targets. A 500-word email that fails on wrong account wastes ~700 tokens on recomposition. The scratchpad eliminates this: compose once, edit by line, send to any target, retry on failure without content loss.

## Test plan

- [x] 115 unit tests for ScratchpadManager (line ops, JSON path, attachments, GC, validation)
- [x] Live tested: compose → email, workspace, doc_create, task_create
- [x] Live tested: import email → edit → doc_create (cross-account: gmail → praecipio)
- [x] Live tested: import Meet transcript → workspace .md → Drive upload (praecipio)
- [x] Live tested: cross-scratchpad copy_lines → email
- [x] Live tested: CSV validation (column count error detection + inline fix)
- [x] Live tested: JSON path editing (json_get, json_set)
- [ ] Sheet write adapter (not live tested — no test spreadsheet)
- [ ] Calendar event adapter (not live tested)
- [ ] Live binding JSON sync (not live tested — needs doc import in JSON mode)

### DIFF
--- a/.claude/ways/api-design/dual-modes/way.md
+++ b/.claude/ways/api-design/dual-modes/way.md
@@ -1,0 +1,25 @@
+---
+description: Offering both readable and lossless modes when there's a fidelity trade-off
+vocabulary: markdown json fidelity lossy lossless readable opaque mode format import export csv rich text plain
+pattern: import|export|format.*mode|markdown.*json|fidelity
+files: src/services/|src/server/
+---
+# Dual Modes at Fidelity Boundaries
+
+When a content format has a genuine trade-off between human readability and structural fidelity, don't force one choice. Offer both modes and let the agent or user pick based on the task.
+
+**The pattern:**
+
+| Mode | Properties | When to use |
+|------|-----------|-------------|
+| Readable (markdown, CSV, plain text) | Lossy but editable, inspectable, cross-service portable | Rewriting, extracting, composing, human review |
+| Lossless (JSON, native API structures) | Full fidelity but opaque, harder to navigate | Targeted edits preserving formatting, round-trip mutations |
+
+**Where this applies in GWS:**
+
+- Google Docs: markdown export vs Docs API JSON
+- Google Sheets: CSV lines vs Sheets API JSON
+- Email: plain text body vs MIME structure
+- Drive files: text extraction vs native format
+
+**When adding a new import/export path**, check whether both modes are feasible. If the lossy mode drops information the agent might need, the lossless mode should exist too — even if it's added later.
--- a/.claude/ways/api-design/full-coverage/way.md
+++ b/.claude/ways/api-design/full-coverage/way.md
@@ -1,0 +1,13 @@
+---
+description: Cross-cutting features should cover all services, not a subset
+vocabulary: cross-cutting coverage services partial complete scope horizontal all seven email docs drive sheets calendar tasks meet scratchpad
+pattern: cross.cutting|all.*service|coverage|scratchpad.*service
+files: src/services/|src/factory/manifest\.yaml
+---
+# Full Coverage for Cross-Cutting Features
+
+When a capability spans multiple GWS services — the scratchpad, batch operations, account management, or anything horizontal — commit to covering all supported services. Partial coverage creates confusion about when the feature applies and when to fall back to direct operations.
+
+- **At design time:** Map the feature to every service. Some mappings may be trivial (calendar events are text, no special adapter needed). Document which services are covered and which are deferred with a reason.
+- **At implementation time:** It's fine to ship incrementally — email and docs first, then sheets, then the rest. But the design should account for all services from the start so later additions don't require rearchitecting.
+- **The anti-pattern:** Building a feature for 2-3 services and calling it done. Agents can't predict which services support the feature and which don't — they'll try it everywhere and hit confusing errors on uncovered services.
--- a/.claude/ways/api-design/late-binding/way.md
+++ b/.claude/ways/api-design/late-binding/way.md
@@ -1,0 +1,15 @@
+---
+description: Decoupling composition from delivery, late-bound targets, deferred binding
+vocabulary: bind target send deliver compose buffer scratchpad decouple recipient dispatch late early creation
+pattern: send|target|deliver|bind|scratchpad
+files: src/services/|src/server/
+---
+# Late-Bound Targets
+
+In this project, content composition and content delivery are separate concerns. When designing operations that create content and then deliver it somewhere:
+
+- **Bind the target at send time, not at creation time.** A piece of content should not know where it's going until the agent explicitly delivers it.
+- **Why:** Retry resilience (failed sends don't lose content), multi-recipient delivery (same content to N targets), cross-service reuse (email content becomes a doc), and simpler agent logic (compose first, decide destination later).
+- **The anti-pattern:** Requiring a destination when the buffer/draft is created. This couples the lifecycle of the content to a single delivery path. If that path fails or changes, the content is stranded.
+
+This applies to the scratchpad (ADR-301) and to any future primitive that holds agent-authored content before delivery.
--- a/.claude/ways/api-design/tool-surface/way.md
+++ b/.claude/ways/api-design/tool-surface/way.md
@@ -1,0 +1,27 @@
+---
+description: Tool response design, what agents see, hiding plumbing, binary references, clean output
+vocabulary: tool response format output agent surface visible opaque plumbing implementation detail binary base64 reference marker attachment blob
+pattern: tool.*response|format.*output|patch\.ts|formatter
+files: src/services/.*patch\.ts|src/factory/
+---
+# Tool Surface Design
+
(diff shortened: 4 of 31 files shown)


aaronsb:
## Code Review: Scratchpad Buffer (ADR-301)

**PR size:** ~4000 lines added (thorough review tier). 857-line manager, 528-line handler, 13 adapter files, 1022-line test suite, ADR.

**Overall assessment:** This is a well-designed, well-tested feature that faithfully implements ADR-301. The architecture is sound — late-bound targets, epoch-based GC, adapter decomposition, and the dual addressing model (line + JSON path) are all good calls. The ADR is one of the better ones I have reviewed: it explains trade-offs clearly and documents rejected alternatives.

That said, there are several findings worth addressing before merge.

---

### Findings

#### 1. SECURITY — Attachment filename collision in email import

**Location:** `src/server/scratchpad/adapters/import-email.ts:74`

**Problem:** Email attachment filenames are used verbatim as workspace filenames via `resolveWorkspacePath(att.filename)`. A malicious email with an attachment named `../../etc/passwd` or `important-existing-file.txt` could overwrite files. While `resolveWorkspacePath` has path traversal protection (resolves and checks the prefix), there is no uniqueness guarantee — a second email import with a file named `report.pdf` silently overwrites the first.

**Why it matters:** Silent data loss for legitimate use; the workspace path safety stops traversal but not collision.

**Suggestion:** Prefix attachment filenames with the scratchpad ID or a hash, e.g., `sp-abc123-report.pdf`, or use a subdirectory per scratchpad. The same concern applies to `import-doc.ts:142` where extracted images are written as `image-1.png`, `image-2.png` — these are globally scoped, not per-scratchpad.

#### 2. SECURITY — Email draft MIME header injection

**Location:** `src/server/scratchpad/adapters/send-email-draft.ts:34-40`

**Problem:** The `to` and `subject` values are interpolated directly into RFC 2822 headers without sanitization:
```typescript
if (to) headers.push(`To: ${to}`);
if (subject) headers.push(`Subject: ${subject}`);
```
A `subject` containing `\r\n` could inject additional headers (e.g., `Bcc: attacker@evil.com`). The `content` also goes directly into the MIME body after the blank-line separator.

**Why it matters:** Header injection is a classic email vulnerability. Even though the LLM is the caller, the parameters ultimately come from user instructions which could be influenced by prompt injection.

**Suggestion:** Strip or reject `\r` and `\n` characters from `to` and `subject` values before building the headers.

#### 3. SOLID — Manager at 857 lines (priority threshold)

**Location:** `src/server/scratchpad/manager.ts`

**Problem:** The file is at 857 lines, crossing the 800-line priority threshold. It combines five distinct responsibilities: buffer CRUD, line operations, JSON path operations, attachment management, validation, and formatting helpers.

**Why it matters:** The JSON path helpers alone (lines 780-857) and the validation functions (lines 678-778) are self-contained — they share no state with the class and could be extracted without any interface changes.

**Suggestion:** Extract to focused modules:
- `validation.ts` — `validate()`, `validateMarkdown()`, `validateJson()`, `validateCsv()`, `countCsvColumns()`
- `json-path.ts` — `parsePath()`, `getByPath()`, `setByPath()`, `deleteByPath()`
- `formatting.ts` — `formatNumberedLines()`, `formatContext()`, `formatRemoveContext()`, `formatSize()`, `normalizeAndSplit()`

This would bring `manager.ts` to ~400 lines focused on buffer state management. The helpers are pure functions — extraction is zero-risk.

#### 4. BUG — Attachment refId assumption in `stripBase64Images`

**Location:** `src/server/scratchpad/adapters/import-doc.ts:146-155`

**Problem:** The function calls `scratchpads.attach()` (which assigns `att-${sp.attachments.size + 1}`) but then independently calculates `const refId = 'att-${i + 1}'` for the marker replacement. If the scratchpad already has attachments before this import (from a prior import into the same buffer, which ADR-301 explicitly supports), the two numbering schemes diverge — the marker will reference the wrong attachment.

**Why it matters:** The ADR specifically says "import appends to the existing buffer (does not replace). Multiple imports can compose into one scratchpad." This breaks that contract.

**Suggestion:** Use the `refId` returned by `scratchpads.attach()` instead of computing it independently:
```typescript
const result = scratchpads.attach(scratchpadId, { ... });
cleaned = cleaned.replace(full, `![${alt}](att:${result.refId} ...)`);
```

#### 5. EDGE CASE — `gc()` mutates Map during iteration

**Location:** `src/server/scratchpad/manager.ts:589-595`

**Problem:** The `gc()` method iterates `this.scratchpads.values()` while calling `this.scratchpads.delete()` inside the loop. While this works correctly in V8's current implementation (Map iteration handles concurrent deletion), it relies on implementation-specific behavior rather than the spec guarantee.

**Suggestion:** Collect IDs to delete into an array first, then delete in a second pass. Minor, but removes the spec-ambiguity concern.

#### 6. DESIGN — Live binding sync silently discards local edits for Docs

**Location:** `src/server/scratchpad/handler.ts:314-342`

**Problem:** The `syncIfBound` function for Docs does not actually push local changes to the API — the comment at lines 325-329 acknowledges this ("Full batchUpdate translation is a future enhancement"). Instead, it re-fetches from the API, which **overwrites the local mutation the user just made**. The user calls `json_set`, the buffer is mutated locally, then `syncIfBound` reloads from the API and replaces the buffer with the old state.

**Why it matters:** This contradicts the ADR's promise that "each mutation is atomic" and "the buffer always reflects the live state." The agent will see a success message for `json_set`, but the change will be silently reverted on the next view.

**Suggestion:** Either (a) remove the reload for Docs until batchUpdate translation is implemented, leaving the local mutation as source of truth, or (b) skip `syncIfBound` for Docs-bound scratchpads and document the limitation. The Sheets path works correctly because it does push changes before reloading. This is the most important finding in this review.

#### 7. TESTING — No handler/adapter tests

**Problem:** The 115-test suite covers `ScratchpadManager` thoroughly (line ops, JSON path, GC, validation, attachments). However, there are no tests for `handler.ts` (operation routing, parameter validation, `syncIfBound` logic) or for any adapter. The handler has ~350 lines of routing and validation logic, and the adapters contain format-specific conversion (CSV parsing, MIME construction, base64 decoding).

**Why it matters:** The handler is where bugs like the sync issue (finding #6) live. The manager tests prove the buffer works; the handler tests would prove the integration works.

**Suggestion:** Add at minimum:
- Handler tests for parameter validation (missing required params, unknown operations)
- Handler tests for `syncIfBound` (mocked `execute`) — this would have caught finding #6
- Adapter unit tests for `parseCsvLine()` and the MIME draft builder

---

### What looks solid

- **Adapter decomposition:** 13 focused adapter files, each under 130 lines, each with a clear interface contract. Good application of Single Responsibility.
- **Error containment:** Every adapter catches at the boundary, preserves the scratchpad, and returns an actionable error message with the scratchpadId. The "buffer survives failed sends" guarantee is consistently implemented.
- **Epoch-based GC:** Much better than wall-clock timeouts for an MCP server. The 100-epoch threshold is reasonable and the rationale in the ADR is convincing.
- **Workspace path safety:** Both `resolveWorkspacePath` (traversal) and `verifyPathSafety` (symlink) are used correctly in the workspace adapter and attach handler.
- **Test coverage of the manager:** 115 tests covering all line operations, JSON path ops, GC lifecycle, validation, CRLF normalization, edge cases (empty buffers, out-of-range). Thorough.
- **ADR quality:** ADR-301 is well-written. The context section identifies real problems with token costs. The decision documents the data model, lifecycle, and trade-offs clearly. The "Alternatives Considered" section is genuine, not pro-forma.

### Considerations

- **Memory cap:** ADR-301 notes "pathological use would need a cap" but there is no limit on scratchpad count or total buffer size. Worth adding a maximum (e.g., 50 scratchpads, 10MB total) to prevent runaway memory from a looping agent.
- **`meet` adapter pagination:** `import-meet.ts:51` uses `pageSize: 100` for transcript entries. Long meetings will truncate. Either paginate or document the limit.
- **`queue_operations` enum:** `src/server/tools.ts:116` adds `manage_scratchpad` to the queue tool enum but is missing `manage_docs`, `manage_sheets`, `manage_tasks`, `manage_workspace`, `manage_meet` from the enum. This may cause schema validation failures for queued operations using those tools. (Could be pre-existing.)

---

PR #80, opened 2026-03-27
feat(workspace): nested directories, move, and mkdir

aaronsb:
### Workspace directory nesting
- Add `sanitizePath()` — path-aware sanitizer that preserves `/` separators while sanitizing each segment individually and rejecting `..`/`.` traversal
- `resolveWorkspacePath()` now supports nested paths like `reports/q1/summary.csv`
- `list` operation is recursive with tree-formatted output; accepts optional `path` param to scope to subdirectory
- `delete` handles directories recursively
- New `move` operation — mv semantics (move and rename), auto-creates destination parent dirs
- New `mkdir` operation — explicit directory creation
- `manage_workspace` added to `queue_operations` tool enum

### Inline image viewing
- Extend `HandlerResponse` with optional MCP content blocks (image type)
- `server.ts` spreads content blocks into MCP `CallToolResult`
- Workspace `read` returns images inline as base64 (up to 5MB)
- Gmail `getAttachment` and Drive `download`/`export` auto-include image blocks
- New `viewAttachment` (gmail) — view image attachment inline without saving to workspace
- New `viewImage` (drive) — view image file inline without saving to workspace
- Scratchpad image preview covered via workspace `read` + drive `viewImage`

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx jest` — 511 tests pass
- [ ] Manual: write nested path, list shows tree, move relocates, mkdir creates, delete removes dir
- [ ] Manual: download image via drive, verify inline image block in response
- [ ] Manual: viewAttachment on email with image attachment
- [ ] Manual: viewImage on drive image file
- [ ] Manual: workspace read on saved image file

### DIFF
--- a/src/__tests__/executor/workspace.test.ts
+++ b/src/__tests__/executor/workspace.test.ts
@@ -1,4 +1,4 @@
-import { validateWorkspaceDir, resolveWorkspacePath, getWorkspaceDir, checkWorkspaceStatus } from '../../executor/workspace.js';
+import { validateWorkspaceDir, resolveWorkspacePath, getWorkspaceDir, checkWorkspaceStatus, sanitizePath } from '../../executor/workspace.js';
 import * as os from 'node:os';
 import * as path from 'node:path';
 
@@ -44,6 +44,45 @@ describe('validateWorkspaceDir', () => {
   });
 });
 
+describe('sanitizePath', () => {
+  it('preserves directory separators', () => {
+    expect(sanitizePath('reports/q1/summary.csv')).toBe(path.join('reports', 'q1', 'summary.csv'));
+  });
+
+  it('sanitizes each segment individually', () => {
+    expect(sanitizePath('reports/<bad>/file.txt')).toBe(path.join('reports', '_bad_', 'file.txt'));
+  });
+
+  it('rejects .. traversal segments', () => {
+    expect(() => sanitizePath('../etc/passwd')).toThrow(/traversal/);
+    expect(() => sanitizePath('reports/../../etc/passwd')).toThrow(/traversal/);
+  });
+
+  it('rejects . segments', () => {
+    expect(() => sanitizePath('./file.txt')).toThrow(/traversal/);
+  });
+
+  it('handles empty input', () => {
+    expect(sanitizePath('')).toBe('unnamed');
+  });
+
+  it('collapses empty segments from double slashes', () => {
+    expect(sanitizePath('reports//q1///file.txt')).toBe(path.join('reports', 'q1', 'file.txt'));
+  });
+
+  it('normalizes backslashes', () => {
+    expect(sanitizePath('reports\\q1\\file.txt')).toBe(path.join('reports', 'q1', 'file.txt'));
+  });
+
+  it('strips leading dots from segments', () => {
+    expect(sanitizePath('.hidden/file.txt')).toBe(path.join('hidden', 'file.txt'));
+  });
+
+  it('handles single filename (no separators)', () => {
+    expect(sanitizePath('report.csv')).toBe('report.csv');
+  });
+});
+
 describe('resolveWorkspacePath', () => {
   const origEnv = process.env.WORKSPACE_DIR;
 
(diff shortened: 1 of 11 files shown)


aaronsb (commented):
## Code Review -- PR #80

**Scope**: ~420 lines across 11 files. Two features: workspace directory nesting and inline image viewing.

---

### Security: sandbox and traversal

The security model is sound overall. `sanitizePath` rejects `..` and `.` segments before `sanitizeFilename` can strip them into something that passes. `resolveWorkspacePath` has a belt-and-suspenders `startsWith` check, and `verifyPathSafety` resolves symlinks with `fs.realpath`. All write/read/delete/move/mkdir paths go through both `resolveWorkspacePath` and `verifyPathSafety`. Good layered defense.

**One gap: `listRecursive` does not call `verifyPathSafety` on individual entries.**

`src/server/handlers/workspace.ts:16-39` -- The function calls `verifyPathSafety` on `listRoot` (line 65), but then walks the tree without verifying each child path. If someone places a symlink inside the workspace pointing outside it, `listRecursive` will follow it through `fs.readdir` + `fs.stat`, leaking file names and sizes from outside the sandbox. The `readdir({ withFileTypes })` call follows symlinks by default.

This is a low-severity info-leak (filenames and sizes, not content), and exploiting it requires an attacker who can already create symlinks inside the workspace. Still, for consistency with the rest of the sandbox model, consider either:
- Calling `verifyPathSafety(path.join(root, rel))` inside the loop before recursing/stat-ing
- Or using `lstat` instead of `stat` and skipping symlinks entirely

**Recursive delete (`fs.rm`) follows symlinks too.** At `workspace.ts:177`, `fs.rm(filePath, { recursive: true })` will follow symlinks inside the directory tree and delete targets outside the sandbox. This is the higher-severity variant of the same issue. Consider passing `{ recursive: true, force: false }` and checking for symlinks before recursing, or at minimum calling `fs.lstat` on children. Node's `fs.rm` does NOT resolve symlinks for the top-level target, but it does follow them for children. Worth a test case.

---

### Correctness and edge cases

**`sanitizePath` encoded traversal bypass** -- `src/executor/workspace.ts:139-154`

The function checks for literal `..` and `.` segments. What about segments like `%2e%2e` or `\x2e\x2e`? These would survive `sanitizeFilename` (which strips control chars and `<>:"|?*` but not `%` or `.`). However, the downstream `path.resolve` + `startsWith` check in `resolveWorkspacePath` catches these because the resolved path would still be inside the workspace. So this is defense-in-depth working correctly -- just noting it for awareness.

**.hidden directory stripping may surprise users** -- `sanitizeFilename` strips leading dots, so `.git/config` becomes `git/config`. The test at line 45-46 documents this, which is good. But users expecting to create `.env` files in the workspace will be silently redirected to `env`. Consider whether this is intentional for directories (it was fine for filenames where hiding is undesirable, but for nested paths it changes semantics).

**`move` operation: no source existence check** -- `workspace.ts:192-210` -- `fs.rename` will throw ENOENT, which bubbles up as-is. The error message won't be user-friendly ("ENOENT: no such file or directory" without mentioning the source path). Minor, but wrapping with context per the error handling way would improve the agent experience.

**`move` operation: cross-device rename** -- `fs.rename` fails with EXDEV if source and destination are on different filesystems (possible if WORKSPACE_DIR is on a different mount). This is an edge case but worth noting. A copy+delete fallback would be more robust.

**`viewImage` temp file name** -- `src/services/drive/patch.ts:129` -- `gws-view-${fileId}-${Date.now()}` in `os.tmpdir()`. The `fileId` comes from user input and is not sanitized before being used as part of a filename in `/tmp`. Drive file IDs are alphanumeric in practice, but if the pattern changes or this is called with a crafted value, it could include path separators. Consider sanitizing `fileId` or using `crypto.randomUUID()` instead.

---

### Missing error handling

**`listRecursive` silently swallows all errors** -- `workspace.ts:22-25` -- The `catch {}` block returns an empty array for ANY error, not just ENOENT. Permission errors, I/O errors, and other unexpected failures will be silently lost. Per the error handling way, only expected operational errors should be caught. Consider catching only ENOENT:

```typescript
} catch (err) {
  if ((err as NodeJS.ErrnoException).code === 'ENOENT') return entries;
  throw err;
}
```

**`buildImageBlockFromFile` double-reads** -- `file-output.ts:73-79` -- It calls `fs.stat` then `fs.readFile`. The workspace `read` handler also calls `fs.stat` before calling `buildImageBlockFromFile`. This means for every image read, there are two `stat` calls. Not a bug, but a minor efficiency note.

---

### Design observations (not blocking)

**SVG dual classification** -- SVG appears in both `TEXT_EXTENSIONS` (line 24) and `IMAGE_MIME_MAP` (line 35). When processing an SVG file, `isTextFile` returns true, so the text path wins in `saveToWorkspace` (the `if/else if` on lines 115-119). But in workspace `read`, the image path is tried first (line 113), so SVGs get base64-encoded as images rather than returned as readable text. This inconsistency could confuse agents. Consider deciding: is SVG text or image in this system?

**Content blocks type typing** -- `server.ts:86-91` -- The content array is typed `Array<Record<string, unknown>>` rather than using a discriminated union. This works but loses type safety. Minor, not blocking.

**`manage_workspace` in `queue_operations`** -- Good addition. Workspace file management operations (list, organize) are natural candidates for batch queuing.

---

### Test coverage

The `sanitizePath` tests (workspace.test.ts) are thorough for the happy path and common traversal attacks. Missing test cases worth adding:
- Symlink inside workspace pointing outside (integration test)
- Recursive delete on directory containing symlink
- `move` to existing destination (overwrite behavior)
- `list` on a path that is a file, not a directory
- `viewImage`/`viewAttachment` with a non-image file (error path)

---

### Summary

The sandbox security model is well-structured with multiple layers. The main actionable items are:

1. **[Medium]** `listRecursive` and recursive `delete` follow symlinks without verification -- potential sandbox escape for listing and deletion
2. **[Low]** `listRecursive` silently swallows all errors, not just ENOENT
3. **[Low]** `viewImage` temp filename uses unsanitized `fileId`
4. **[Low]** SVG is classified as both text and image with inconsistent handling between paths

None of these are merge-blockers for a workspace sandbox where the threat model assumes the agent is the primary actor. Items 1 and 2 are worth addressing before or shortly after merge.

---

PR #96, opened 2026-04-08
feat(coverage): build-time gws CLI coverage analysis

aaronsb:
## Summary
- New `src/coverage/` module that compares the curated manifest against the full gws CLI surface
- Reuses factory manifest parser and `resolveGwsBinary()` from the executor
- Discovers 12 eligible services via `gws --help` and `gws schema`
- Produces terminal report with per-service coverage bars, new operations, and parameter gaps
- Baseline tracking (`coverage-baseline.json`) for drift detection and intentional exclusions

## Usage
```bash
make coverage        # print coverage report
make coverage-update # also update baseline file
```

## Initial findings
- **62/337 operations (18%)** covered across 12 services
- 5 uncovered but eligible services: slides, people, chat, keep, events
- Parameter gaps on covered operations (e.g., `supportsAllDrives` showing as gap on Drive ops where it's in defaults but not user-facing params)

Implements ADR-100.

## Test plan
- [x] `make build` succeeds
- [x] `make coverage` prints formatted report
- [x] `make coverage-update` writes `coverage-baseline.json`
- [x] Re-run after update shows no "New since baseline" entries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,8 +33,9 @@ Thumbs.db
 logs/
 *.log
 
-# Test coverage
+# Test coverage (jest output, not src/coverage/)
 coverage/
+!src/coverage/
 commit-message.txt
 discovered-manifest.yaml
 
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .DEFAULT_GOAL := help
 .PHONY: help test test-all test-unit test-integration build clean typecheck \
         manifest-discover manifest-diff manifest-lint \
+        coverage coverage-update \
         mcpb mcpb-all version-sync publish-all \
         release-patch release-minor release-major check
 
@@ -65,6 +66,14 @@ manifest-lint: ## Validate manifest YAML syntax and structure
 		} \
 		console.log('  Total: ' + ops + ' operations across ' + Object.keys(m.services).length + ' services'); \
 	"
+
+# --- Coverage analysis ---
+
+coverage: build ## Analyze gws CLI coverage vs curated manifest
+	node build/coverage/analyze.js
+
+coverage-update: build ## Update coverage baseline from current gws surface
+	node build/coverage/analyze.js --update
 
 # --- MCPB packaging ---
 
--- a/coverage-baseline.json
+++ b/coverage-baseline.json
@@ -1,0 +1,1153 @@
+{
+  "gwsVersion": "gws 0.22.5",
+  "generatedAt": "2026-04-08T15:22:03.016Z",
+  "services": {
+    "drive": {
+      "operations": {
+        "files.copy": {
+          "status": "covered",
+          "params": {
+            "ignoreDefaultVisibility": "gap",
+            "includeLabels": "gap",
+            "includePermissionsForView": "gap",
+            "keepRevisionForever": "gap",
+            "ocrLanguage": "gap",
+            "supportsAllDrives": "gap"
+          }
+        },
+        "files.delete": {
+          "status": "covered",
+          "params": {
+            "supportsAllDrives": "gap"
+          }
+        },
+        "files.get": {
+          "status": "covered",
+          "params": {
+            "acknowledgeAbuse": "gap",
+            "includeLabels": "gap",
+            "includePermissionsForView": "gap",
+            "supportsAllDrives": "gap"
+          }
+        },
+        "files.list": {
+          "status": "covered",
+          "params": {
+            "corpora": "gap",
+            "driveId": "gap",
+            "includeItemsFromAllDrives": "gap",
+            "includeLabels": "gap",
+            "includePermissionsForView": "gap",
+            "orderBy": "gap",
+            "pageToken": "gap",
+            "spaces": "gap",
+            "supportsAllDrives": "gap"
+          }
+        },
+        "permissions.create": {
+          "status": "covered",
+          "params": {
+            "emailMessage": "gap",
+            "moveToNewOwnersRoot": "gap",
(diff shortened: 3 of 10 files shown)


aaronsb:
## Coverage Analysis Tool — Code Review

**Scope**: ~400 lines of new TypeScript across 6 modules, plus Makefile targets, baseline file, and docs update.

**Overall assessment**: Clean, well-decomposed module with good separation of concerns. The architecture follows SRP well — each file has one job (types, discovery, comparison, baseline, reporting, orchestration). Reuse of `loadManifest()` and `resolveGwsBinary()` is the right call. A few items worth addressing, mostly around robustness of help-output parsing and a gap in baseline generation.

---

### Issue: Schema Path Double-Prefix (Medium — Correctness)

**Location**: `src/coverage/discover.ts:128`

**Problem**: The schema path is constructed as `${service}.${parentPath}.${method}`, but `parentPath` is already relative to the service (e.g., `files`). If `gws schema` expects `drive.files.copy`, this is correct. But for nested sub-resources, `parentPath` accumulates (e.g., `permissions.list` under `files` would become `files.permissions.list`), and the schema call becomes `drive.files.permissions.list`. This is *probably* correct for the gws schema command, but it would be good to confirm nested sub-resources resolve properly, since no test validates this path.

**Suggestion**: Worth a manual smoke test with a nested resource (e.g., `gws schema drive.files.permissions.list`) to confirm the schema command accepts the same dot-path structure the discovery builds.

---

### Issue: Help-Output Parsing Fragility (Medium — Robustness)

**Location**: `src/coverage/discover.ts:69-84` (`parseResourceHelp`)

**Problem**: The method regex `^(\w+)\s+\S` will match *any* line that starts with a word followed by whitespace and a non-space character. This includes flag lines like `--verbose  Enable verbose output` (won't match because of `--`), but could match section headers or other help text that happens to look like `word  description`. The function skips `Usage:` and `FLAGS:` lines, but doesn't skip other common sections like `COMMANDS:`, `DESCRIPTION:`, or the binary name line.

**Suggestion**: Consider adding a section-tracking state machine, or at minimum skip lines that match common gws help boilerplate patterns (e.g., lines starting with uppercase words followed by `:`, or the command name echo). Alternatively, if gws help output has a consistent structure, restrict parsing to only the "commands" section.

---

### Issue: Baseline Generation Loses Covered Operations Without Param Gaps (Medium — Correctness)

**Location**: `src/coverage/baseline.ts:40-65` (`generateBaseline`)

**Problem**: The baseline generation iterates `svc.paramGaps` to find covered ops (line 40), then marks `newOps` as gaps (line 61), and preserves `excluded` entries from the existing baseline (line 54). But **covered operations that have zero param gaps are never written to the baseline file**. This means the baseline only tracks: (1) covered ops *with* param gaps, (2) excluded ops, and (3) new gaps. A covered op with no gaps simply disappears from the baseline.

This is fine if the baseline is only meant to track *problems and exclusions*, but it means the baseline cannot reconstruct the full picture of what was covered. The next run's "new since baseline" detection (`!wasInBaseline` in compare.ts:103) won't flag these operations as new (because they're covered, not gaps), so the functional impact is limited. But the baseline file's incompleteness is surprising — a reader might expect it to be a complete snapshot.

**Suggestion**: Either (a) document that the baseline intentionally only tracks ops with gaps/exclusions (a comment would suffice), or (b) emit a `{ status: 'covered' }` entry for all covered ops so the baseline is a complete picture. Option (b) is more robust for future features like trend tracking.

---

### Issue: No Unit Tests (Low — Test Coverage)

**Problem**: No test files exist for any of the coverage modules. The parsing logic in `parseServiceHelp`, `parseResourceHelp`, `parseSchemaParams`, and the comparison logic in `compareSurfaces` are all purely functional and highly testable without mocking the gws binary.

**Suggestion**: At minimum, unit tests for:
- `parseServiceHelp` with sample help output (resources + helpers detected correctly)
- `parseResourceHelp` with edge cases (empty output, section headers that look like methods)
- `compareParams` with various manifest/gws param combinations
- `generateBaseline` round-trip (generate -> load -> re-generate preserves exclusions)

These are pure functions — they take strings/objects and return structured data. Easy wins.

---

### Issue: resolveGwsBinary Called Per Invocation (Low — Performance)

**Location**: `src/coverage/discover.ts:17,27`

**Problem**: Both `gws()` and `gwsSafe()` call `resolveGwsBinary()` on every invocation. For a full surface discovery of 12 services, each with multiple resources and methods, this could mean hundreds of calls to `resolveGwsBinary()`. If that function does filesystem lookups, it's wasteful.

**Suggestion**: Resolve once at module level or at the start of `discoverSurface()` and pass the binary path to `gws`/`gwsSafe`.

---

### Observations (No Action Needed)

- **Good**: The `SKIP_PARAMS` set correctly filters out auth/transport params that would create noise in the coverage report.
- **Good**: The `gwsSafe` wrapper with swallowed stderr and null return is appropriate for optional discovery — a missing resource shouldn't crash the tool.
- **Good**: Report formatting uses stderr for progress and stdout for output, making it pipeable (`make coverage --json | jq`).
- **Good**: The `!src/coverage/` gitignore negation is the right fix for the naming collision with jest's `coverage/` output.
- **Good**: The Makefile targets depend on `build`, ensuring TypeScript is compiled first.

---

**Summary**: Solid build tool with clean module boundaries. The baseline incompleteness (covered ops without gaps not recorded) is the most impactful finding — worth either fixing or documenting. The help-parsing fragility is a known trade-off of scraping CLI output; adding tests with representative samples would make it safe to iterate on.

PR #97, opened 2026-04-08
feat(docs,drive): comments and inline editing

aaronsb:
## Summary
Adds document comments and inline editing operations, following the inject-after-factory pattern.

### Drive comments (via Drive API v3)
- **listComments** / **getComment** — manifest-driven with `fields` defaults
- **addComment** — custom handler, supports optional `quotedText` anchor
- **resolveComment** — custom handler to resolve/reopen threads
- **replyToComment** — custom handler for comment replies

### Docs inline editing (via `documents.batchUpdate`)
- **insertText** — insert at character index position
- **replaceText** — find and replace with occurrence count feedback

### Architecture
- New `src/services/docs/patch.ts` for batchUpdate-based operations
- Drive comment handlers added to existing `src/services/drive/patch.ts`
- All comment operations include `supportsAllDrives: true`

Fixes #92

## Test plan
- [ ] `make build` succeeds
- [ ] `listComments` returns comments on a document
- [ ] `addComment` creates a comment (with and without `quotedText` anchor)
- [ ] `resolveComment` resolves and reopens a comment
- [ ] `replyToComment` adds a reply to a comment thread
- [ ] `insertText` inserts text at a given index
- [ ] `replaceText` replaces text and reports occurrence count

### DIFF
--- a/src/factory/manifest.yaml
+++ b/src/factory/manifest.yaml
@@ -456,7 +456,7 @@ services:
 
   drive:
     tool_name: manage_drive
-    description: "Search, upload, download, share, or manage files in Google Drive."
+    description: "Search, upload, download, share, comment on, or manage files in Google Drive."
     requires_email: true
     gws_service: drive
     operations:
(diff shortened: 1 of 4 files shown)


aaronsb:
## Review: feat(docs,drive): comments and inline editing

**Size**: ~305 lines added across 4 files. Clean scope.

### What this does well

- **Architecture is sound.** Comments go through Drive API, inline editing through Docs batchUpdate — correct API surface for each. The inject-after-factory pattern with `--json` for request bodies is well-established in this codebase and applied consistently here.
- **Manifest params match what handlers consume.** Each custom handler destructures exactly the params declared in its manifest entry — no orphaned or missing params.
- **batchUpdate request bodies are correct.** `InsertTextRequest` with `location.index` and `ReplaceAllTextRequest` with `containsText` + `replaceText` match the Docs API v1 spec.
- **Validation and type coercion** are handled properly — `requireString` for required fields, `Number()` with `isFinite` guard on `index`, boolean defaulting via `!== false`.
- **Occurrence count extraction** from `replaceAllText` reply is well done — defensive with fallback to 0.
- **`supportsAllDrives: true`** applied everywhere, consistent with all existing Drive operations.

### Issues to address

#### 1. `resolveComment` — `content: ''` is suspicious

**Location**: `src/services/drive/patch.ts:234-235`

The Drive API `comments.update` requires `content` in the request body, but setting it to empty string will **overwrite the original comment text** to blank. The inline comment says "content is required but preserved server-side" — this is incorrect. Drive API will store whatever you send.

**Suggestion**: Fetch the current comment first to preserve its content, or simply omit the `content` field (the API may accept a body with only `resolved: true` — worth testing). If `content` must be sent, do a `comments.get` first and echo it back.

#### 2. `resolveComment` — asymmetric body for resolve vs reopen

**Location**: `src/services/drive/patch.ts:233-236`

When `resolved` is `true`, the body includes `{ content: '', resolved: true }`. When `resolved` is `false`, the body is just `{ resolved: false }` (the `if (resolved)` guard means `content` is never set for reopen). This asymmetry is a sign the `content: ''` approach needs rethinking — the behavior differs depending on direction.

#### 3. `insertText` — integer validation incomplete

**Location**: `src/services/docs/patch.ts:19-21`

The validation checks `index >= 1` but does not check that it is an **integer**. A fractional index like `1.5` passes the check but is not meaningful for the Docs API (character indices are integers). Add `Math.floor(index) !== index` or use `Number.isInteger()`.

**Suggestion**:
```typescript
if (!Number.isInteger(index) || index < 1) {
  throw new Error('index must be a positive integer (1 = start of document body)');
}
```

### Minor observations (non-blocking)

- **`replaceText` `matchCase` default**: The manifest says `default: true` and the handler defaults via `!== false`. These agree, but the manifest `default` is informational only (not enforced by the factory) — just noting for awareness.
- **`listComments` pagination**: The manifest includes `nextPageToken` in fields, which is good. No `pageToken` param is declared, so the caller cannot paginate manually. Probably fine for now since the factory's `list` type may handle this, but worth confirming when testing with files that have many comments.
- **`quotedFileContent` anchor limitation**: The description says "anchored to first occurrence" — this is a Drive API behavior, not something the handler controls. Accurate documentation choice.
- **No tests**: PR body has a manual test plan. Given these are thin wrappers around `execute()`, integration tests are the right level — just flagging for the record.

### Verdict

The `resolveComment` content-blanking issue (item 1) is the main concern — it could destroy comment text in production. Worth testing or fixing before merge. Items 2 and 3 are quick fixes. Everything else is solid work.

PR #99, opened 2026-04-09
fix(calendar): freebusy, agenda params, calendarId context (#98)

### LINKED ISSUE #98: Calendar: freebusy timeMin not passed to API + get fails on shared calendar events without calendarId
### Expected
Calling `manage_calendar` with `operation: "freebusy"` and `timeMin`/`timeMax` should query the Google Calendar FreeBusy API for the specified time range and return busy blocks for the requested attendees.

### Actual
Always returns `"Missing timeMin parameter."` regardless of which time parameters are provided.

### Reproduction
All of the following fail with the same error:

```json
// Attempt 1: timeMin/timeMax only
{"operation": "freebusy", "email": "user@example.com", "attendees": "colleague@example.com", "timeMin": "2026-04-09T08:00:00-05:00", "timeMax": "2026-04-09T17:00:00-05:00"}

// Attempt 2: start/end only
{"operation": "freebusy", "email": "user@example.com", "attendees": "colleague@example.com", "start": "2026-04-09T08:00:00-05:00", "end": "2026-04-09T17:00:00-05:00"}

// Attempt 3: both pairs simultaneously
{"operation": "freebusy", "email": "user@example.com", "attendees": "colleague@example.com", "start": "2026-04-09T08:00:00-05:00", "end": "2026-04-09T17:00:00-05:00", "timeMin": "2026-04-09T08:00:00-05:00", "timeMax": "2026-04-09T17:00:00-05:00"}
```

Error in all cases:
```json
{"error": "Missing timeMin parameter.", "exitCode": 1, "reason": "required"}
```

Likely cause: the freebusy operation handler isn't mapping the incoming tool parameters to the Google Calendar API's `timeMin`/`timeMax` fields in the request body.

---

### Expected
Event IDs returned by `list` (scoped to a shared calendar via `calendarId`) should be fetchable via `get` using the same event ID. When `calendarId` was used to discover the event, it should also work for retrieval — or the output should make the required calendarId obvious.

### Actual
`get` with a valid event ID from a shared calendar returns `Not Found` when `calendarId` is omitted (defaults to primary). Works correctly when `calendarId` is explicitly set to the shared calendar.

### Reproduction

```json
// Step 1: List events on shared calendar — works fine, returns event IDs
{"operation": "list", "email": "user@example.com", "calendarId": "shared@example.com", "timeMin": "2026-04-09T00:00:00-05:00", "timeMax": "2026-04-10T00:00:00-05:00"}
// Returns event ID: "abc123"

// Step 2a: Get event without calendarId — FAILS
{"operation": "get", "email": "user@example.com", "eventId": "abc123"}
// Error: {"error": "Not Found"}

// Step 2b: Get event WITH calendarId — WORKS
{"operation": "get", "email": "user@example.com", "calendarId": "shared@example.com", "eventId": "abc123"}
// Returns full event details
```

### Suggestion
The `agenda` operation surfaces events from all calendars and includes the calendar name in its output, but doesn't include the `calendarId` needed for follow-up `get` calls. A couple of options:
- Include `calendarId` in agenda/list output so consumers know which calendar to target
- Have `get` search across accessible calendars when the event isn't found on primary

aaronsb:
## Summary

- **freebusy**: Fixed — was sending POST body as `--params` (query), now uses `--json` (request body). Added `attendees` and `calendarId` params with formatted busy/free output.
- **agenda**: Exposed `days`, `tomorrow`, `week`, `calendarId` filter params from the `+agenda` helper. Includes `calendarId` per event in refs for follow-up `get` calls.
- **list**: Enriched output with `calendarId` in refs and header when querying shared calendars.
- **create**: Added `--meet` flag for Google Meet video conference links.
- **get**: Updated description noting `calendarId` is required for shared calendar events.

Closes #98

## Test plan
- [ ] `freebusy` with timeMin/timeMax returns busy blocks (was always failing before)
- [ ] `freebusy` with attendees checks multiple calendars
- [ ] `agenda --week` and `agenda --days 3` return multi-day results
- [ ] `list` on shared calendar includes calendarId in refs
- [ ] `get` with calendarId from list refs succeeds on shared calendar events
- [ ] `create` with `meet: true` produces event with Meet link

### DIFF
--- a/src/__tests__/factory/calendar-patch.test.ts
+++ b/src/__tests__/factory/calendar-patch.test.ts
@@ -1,0 +1,286 @@
+/**
+ * Tests for the calendar service patch — custom handlers and formatters
+ * that extend the factory-generated handler.
+ */
+
+import {
+  mockExecute, mockGwsResponse,
+  calendarAgendaResponse, calendarEventsListResponse, calendarInsertResponse,
+  calendarFreeBusyResponse, calendarFreeBusyErrorResponse,
+} from '../server/handlers/__mocks__/executor.js';
+import { calendarPatch } from '../../services/calendar/patch.js';
+import type { PatchContext } from '../../factory/types.js';
+
+function ctx(overrides: Partial<PatchContext> = {}): PatchContext {
+  return {
+    operation: 'list',
+    params: {},
+    account: 'user@test.com',
+    ...overrides,
+  };
+}
+
+describe('calendarPatch', () => {
+  beforeEach(() => mockExecute.mockReset());
+
+  describe('formatList (events)', () => {
+    it('defaults calendarId to primary in refs', () => {
+      const result = calendarPatch.formatList!(calendarEventsListResponse, ctx({ operation: 'list' }));
+      expect(result.refs.calendarId).toBe('primary');
+    });
+
+    it('enriches refs with calendarId for shared calendars', () => {
+      const result = calendarPatch.formatList!(
+        calendarEventsListResponse,
+        ctx({ operation: 'list', params: { calendarId: 'shared@test.com' } }),
+      );
+      expect(result.refs.calendarId).toBe('shared@test.com');
+    });
+
+    it('adds calendar hint to output header for non-primary calendars', () => {
+      const result = calendarPatch.formatList!(
+        calendarEventsListResponse,
+        ctx({ operation: 'list', params: { calendarId: 'shared@test.com' } }),
+      );
+      expect(result.text).toContain('calendar: shared@test.com');
+    });
+
+    it('does not add hint for primary calendar', () => {
+      const result = calendarPatch.formatList!(calendarEventsListResponse, ctx({ operation: 'list' }));
+      expect(result.text).not.toContain('calendar:');
+    });
+  });
+
+  describe('agenda custom handler', () => {
+    it('defaults to --today flag', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse(calendarAgendaResponse));
+      await calendarPatch.customHandlers!.agenda({}, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args).toContain('--today');
+    });
+
+    it('passes --week flag', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse(calendarAgendaResponse));
+      await calendarPatch.customHandlers!.agenda({ week: true }, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args).toContain('--week');
+      expect(args).not.toContain('--today');
+    });
+
+    it('passes --tomorrow flag', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse(calendarAgendaResponse));
+      await calendarPatch.customHandlers!.agenda({ tomorrow: true }, 'user@test.com');
+
+      expect(mockExecute.mock.calls[0][0]).toContain('--tomorrow');
+    });
+
+    it('passes --days N', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse(calendarAgendaResponse));
+      await calendarPatch.customHandlers!.agenda({ days: 3 }, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args).toContain('--days');
+      expect(args).toContain('3');
+    });
+
+    it('passes --calendar filter', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse(calendarAgendaResponse));
+      await calendarPatch.customHandlers!.agenda({ calendarId: 'work@test.com' }, 'user@test.com');
+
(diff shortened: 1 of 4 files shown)


aaronsb:
## Code Review: Calendar Patch Enhancements

**PR size**: ~146 additions, 22 deletions across 2 files — standard review.

---

### 1. Freebusy: --json body construction (patch.ts:166-188) -- Looks Correct

The core fix is sound. The Google Calendar FreeBusy API expects a POST body with `{ timeMin, timeMax, items: [{ id: "email" }] }`, and the old factory path was incorrectly putting these into `--params` (query string). The custom handler correctly:

- Uses `requireString` for timeMin/timeMax (validated, will throw on missing)
- Always includes the user's own calendar as `{ id: account }`
- Deduplicates calendarId entries against existing items (`items.some(i => i.id === id)`)
- Uses `--json` for the POST body, matching the established pattern (docs/patch.ts, drive/patch.ts, gmail/patch.ts all do the same)

**One concern**: The `attendees` param does **not** get the same deduplication that `calendarId` gets. If a user passes their own email in `attendees`, it would appear twice in the items array. This is harmless (the API deduplicates on its end and returns one entry per calendar), but it is inconsistent within the function — consider applying the same `items.some()` guard to attendees for symmetry, or adding a comment explaining why it is not needed.

### 2. formatFreeBusy response shape (patch.ts:66-96) -- One Risk

The formatter reads `data.calendars` which matches the [FreeBusy API response](https://developers.google.com/calendar/api/v3/reference/freebusy/query#response). This is correct.

**Edge case — empty calendars object**: If the API returns `{ calendars: {} }` (possible if no items matched or permissions denied), the function produces only `## Availability\n` with no entries, plus next steps. The output is technically valid but could be confusing. A minor improvement would be adding a "No calendars returned" fallback when `Object.keys(calendars).length === 0`.

**Edge case — API error blocks**: The FreeBusy API can return per-calendar `errors` arrays alongside `busy` arrays (e.g., `{ "calendars": { "x@y.com": { "errors": [{"domain": "calendar", "reason": "notFound"}], "busy": [] } } }`). The formatter currently ignores these — errors would show as "Free for entire range" which is misleading. This is a pre-existing gap (not introduced by this PR), but worth noting as a follow-up.

### 3. CalendarId enrichment in list/agenda (patch.ts:47-62, 99-116) -- Reliable with Caveat

**formatEventListWithCalendar** (line 47-62): Clean approach. Spreads existing refs and adds calendarId. The regex replacement `^## Events` uses start-of-string anchor which is correct since `formatEventList` produces `## Events (N)` at the beginning.

**formatAgenda** (line 99-116): The calendarId-per-event extraction at line 112 uses a sensible fallback chain: `e.calendarId ?? organizer.email`. This is the right heuristic — the `+agenda` helper aggregates across calendars, and Google's event objects include `organizer.email` which maps to the calendar that owns the event.

**Minor concern**: If `data` is a string (line 103 check), the `events` extraction from `raw.events` on line 102 would yield an empty array. The text would be the raw string, but refs would show `count: 0, events: []`. This is a graceful degradation, not a bug, but it means refs and text could be inconsistent if the `+agenda` helper returns raw text instead of structured JSON.

### 4. Agenda day-range params (patch.ts:150-163, manifest.yaml:320-332)

The mutually exclusive flag handling (`tomorrow` > `week` > `days` > default `--today`) is clean. The `else` chain ensures only one range flag is sent.

**One question**: The manifest declares `days` as `type: number` but the handler does `String(params.days)`. If an LLM sends `days: "3"` (a string), this still works. If it sends `days: 0`, it would be falsy and fall through to `--today`. That seems intentionally correct (0 days = today), but `days: -1` would pass through as `--days -1` — presumably the gws CLI would reject that, so no real risk.

### 5. Manifest param definitions (manifest.yaml) -- Clean

- `agenda` params are well-described and correctly do not have `cli_args` mappings since the custom handler builds args directly
- `freebusy` added `attendees` and `calendarId` — no `cli_args` needed since the custom handler constructs `--json`
- `create` added `meet: boolean` — note there is no `cli_args` mapping for `meet`, which is correct since the custom handler uses `args.push('--meet')` directly
- Updated descriptions on `get` and `freebusy` are accurate and helpful for LLM tool selection

### 6. Testing

No new test files in this PR. The existing `calendar.test.ts` likely does not cover:
- `formatFreeBusy` with various response shapes (empty, errors, multi-calendar)
- `formatAgenda` calendarId extraction
- `formatEventListWithCalendar` enrichment
- Freebusy custom handler item deduplication

Given this fixes a previously-always-failing operation, adding at least a unit test for `formatFreeBusy` would be valuable insurance against regressions.

---

### Summary

The core fix is correct and well-structured — freebusy was fundamentally broken before this change, and the `--json` approach follows established codebase patterns. The calendarId enrichment solves a real usability gap for shared calendars. The code is clean, stays under complexity thresholds, and respects the factory/patch separation.

**Suggested follow-ups** (non-blocking):
1. Add deduplication for attendees in freebusy handler (consistency)
2. Surface per-calendar `errors` from the FreeBusy API response
3. Unit tests for the new formatter functions

PR #101, opened 2026-04-11
fix(drive): create parent directories before export/download write

### LINKED ISSUE #100: manage_drive export operation fails with 'Failed to create output file' on valid paths
## Bug

The `manage_drive` tool's `export` operation consistently fails with:

```
{
  "error": "Failed to create output file: No such file or directory (os error 2)",
  "exitCode": 5,
  "reason": "internalError",
  "stderr": "error: Failed to create output file"
}
```

## Reproduction

1. Use `manage_drive` with `operation: "export"` on a Google Doc
2. Provide any valid `outputPath` — tried multiple existing directories:
   - `/tmp/revenue_analysis.txt`
   - `/home/user/Projects/.../<file>.txt`
   - `/home/user/.local/share/google-workspace-mcp/workspace/<file>.txt`
   - Even pre-created the file first — same error
3. All paths are writable and verified to exist

## Expected behavior

Export should write the file to the specified path.

## Workaround

Using the `gws` CLI directly works fine:

```bash
gws drive files export --params '{"fileId":"<id>","mimeType":"text/plain"}' -o /tmp/output.txt
```

This succeeds and writes the file correctly.

## Environment

- google-workspace-mcp installed via npx (`@aaronsb/google-workspace-mcp`)
- Linux (Arch)
- Claude Code MCP integration

aaronsb:
## Summary
- Export and download custom handlers resolve `outputPath` into the workspace sandbox via `resolveWorkspacePath()`, which can produce nested paths (e.g. user passes `/tmp/file.txt` → becomes `workspace/tmp/file.txt`)
- Neither handler created parent directories before invoking `gws --output`, causing "Failed to create output file: No such file or directory" errors
- Added `fs.mkdir(path.dirname(outputPath), { recursive: true })` before the `execute()` call in both handlers

Closes #100

## Test plan
- [x] New `drive-patch.test.ts` with 3 tests covering subdirectory creation for export and download
- [x] Full test suite passes (548 tests, 34 suites)
- [ ] Manual: export a Google Doc via `manage_drive` with default outputPath
- [ ] Manual: export with explicit nested `outputPath` like `reports/doc.txt`

### DIFF
--- a/src/__tests__/factory/drive-patch.test.ts
+++ b/src/__tests__/factory/drive-patch.test.ts
@@ -36,6 +36,8 @@ const mockExecute = execute as jest.MockedFunction<typeof execute>;
 describe('drivePatch custom handlers', () => {
   beforeEach(async () => {
     mockExecute.mockReset();
+    // Start with a clean workspace — only the root dir exists
+    await fs.rm(tmpWorkspace, { recursive: true, force: true });
     await fs.mkdir(tmpWorkspace, { recursive: true });
   });
 
@@ -45,18 +47,20 @@ describe('drivePatch custom handlers', () => {
 
   describe('export', () => {
     it('creates parent directories before calling gws', async () => {
-      // First call: get file metadata; second call: export
-      mockExecute
-        .mockResolvedValueOnce({ success: true, data: { name: 'Report.gdoc' }, stderr: '' })
-        .mockResolvedValueOnce({ success: true, data: {}, stderr: '' });
-
-      // Write a dummy file so readWorkspaceFile finds something
       const outputPath = path.join(tmpWorkspace, 'subdir', 'Report.txt');
-      await fs.mkdir(path.dirname(outputPath), { recursive: true });
-      await fs.writeFile(outputPath, 'exported content');
 
       const { resolveWorkspacePath } = require('../../executor/workspace.js');
       (resolveWorkspacePath as jest.Mock).mockReturnValue(outputPath);
+
+      // First call: get file metadata
+      // Second call: export — simulate gws writing the file
+      mockExecute
+        .mockResolvedValueOnce({ success: true, data: { name: 'Report.gdoc' }, stderr: '' })
+        .mockImplementationOnce(async () => {
+          // gws would write the file here — parent dir must already exist
+          await fs.writeFile(outputPath, 'exported content');
+          return { success: true, data: {}, stderr: '' };
+        });
 
       const handler = drivePatch.customHandlers!.export!;
       await handler(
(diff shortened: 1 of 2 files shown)


aaronsb:
## Review: fix(drive): create parent directories before export/download write

**Size**: Small (6 lines production, 123 lines test) -- focused and appropriate.

### 1. Correctness -- Does the fix solve the problem?

Yes. The root cause is clear: `resolveWorkspacePath()` can produce nested paths (e.g. user passes `reports/doc.txt` which becomes `workspace/reports/doc.txt`), but `ensureWorkspaceDir()` only creates the workspace root. The two `fs.mkdir` calls with `{ recursive: true }` at lines 79 and 178 of `patch.ts` correctly handle the gap.

The placement is also correct -- `mkdir` runs *after* `verifyPathSafety()` (which allows ENOENT for new files) and *before* `execute()`. This means the security check still runs before any filesystem mutation.

### 2. Security -- Path traversal / sandbox escape risk

**No new risk introduced.** The defense-in-depth chain is solid:

1. `resolveWorkspacePath()` calls `sanitizePath()` which rejects `..` and `.` segments, strips dangerous characters, and verifies the resolved path stays inside the workspace
2. `verifyPathSafety()` checks symlink resolution to catch post-creation escapes
3. Only *then* does `mkdir` run on the already-validated `path.dirname(outputPath)`

Since `outputPath` has already been confirmed to be within the workspace sandbox before `mkdir` executes, the recursive directory creation cannot escape the workspace. Good sequencing.

### 3. Test quality

The tests validate the right things but have a structural weakness worth noting:

**What works well:**
- Tests cover both `export` and `download` handlers
- Tests cover nested paths and flat filenames
- Tests verify `--output` arg points to the correct nested path
- Proper cleanup with `afterAll`

**Weakness -- tests don't actually prove `mkdir` is the thing creating directories:**

The tests manually call `fs.mkdir(path.dirname(outputPath))` *before* invoking the handler (lines 55-56, 106), then assert the directory exists afterward. This proves the handler doesn't *crash* when directories exist, but it doesn't prove the handler *creates* them when they don't exist.

A stronger test would:
1. Set up `resolveWorkspacePath` to return a nested path where the parent does NOT exist
2. Let the handler's `mkdir` create it
3. Assert the parent directory was created

As-is, if you removed the `mkdir` from the production code, these tests would still pass (the test setup creates the directories). This is a meaningful gap -- consider adding a test that omits the manual `mkdir` and relies solely on the handler to create the subdirectory.

### 4. Edge cases

**Covered:**
- Nested subdirectory path (`subdir/Report.txt`, `images/photo.png`)
- Flat filename with no subdirectory (`Doc.pdf`)

**Not covered but low-risk:**
- Deeply nested paths (`a/b/c/d/file.txt`) -- `{ recursive: true }` handles this by design
- `outputPath` where dirname is the workspace root itself -- `mkdir` on an existing dir with `recursive: true` is a no-op, so this is safe

**One observation about `viewImage`:** The `viewImage` handler (line 104) writes to `os.tmpdir()` instead of the workspace, so it doesn't need this fix. Correct that it was left alone.

### 5. SOLID compliance

Clean. The fix follows the existing pattern -- both handlers already had the `ensureWorkspaceDir` / `resolveWorkspacePath` / `verifyPathSafety` sequence, and the `mkdir` slots in naturally. No new responsibilities added to the module, no interface changes. The file at 375 lines is well within size limits.

### Summary

The production fix is correct, well-placed, and introduces no security regression. The test gap (tests pre-create directories instead of verifying the handler creates them) is the only substantive finding. Consider adding one test that starts with a non-existent parent directory to close that gap before merge.

---

PR #103, opened 2026-04-16
fix(sheets): surface values + expand manage_sheets to 13 ops (#90)

aaronsb:
## Summary

Fixes #90 and expands `manage_sheets` from 6 to 13 agent-facing operations. Sheets tool coverage on the gws surface moves from **32% → 47%** (6/19 → 9/19 unique resources; the 5 batchUpdate-based tab ops share one endpoint). Also fixes two latent bugs discovered during the live-test session.

### #90 — read/write bug fix (`19b5303`)
The generic detail formatter in `src/factory/defaults.ts` skipped any object/array field, so `read`/`getValues` dropped the `values` 2D array and `get` dropped the `sheets` list. `updateValues` declared only `spreadsheetId`/`range` with no body wiring, so writes never sent data. New `src/services/sheets/patch.ts`:
- `formatDetail` renders `values` as a pipe-delimited markdown table and shows sheet tabs with dimensions for `get`
- `formatAction` surfaces `create`/`append`/`clearValues` response fields
- `customHandlers.updateValues` assembles the `{values}` body via `--json`, accepts `values` (CSV) or `jsonValues` (JSON 2D array) plus `valueInputOption`

### Tab management (`b0ee618`)
Five new operations, all via `spreadsheets.batchUpdate` with a single-request body:
- `addSheet` — add a tab (optional `title`/`index`/`rowCount`/`columnCount`)
- `renameSheet` — rename a tab by `sheetId`
- `deleteSheet` — remove a tab by `sheetId`
- `duplicateSheet` — clone a tab within the same spreadsheet
- `renameSpreadsheet` — rename the document itself

Plus two more that fit the factory path directly:
- `clearValues` — clear a range (`spreadsheets.values.clear`)
- `copySheetTo` — copy a tab into another spreadsheet (`spreadsheets.sheets.copyTo`)

### `clearValues` formatter polish (`2ee9ff7`)
Surfaces `clearedRange` instead of a generic "Operation completed."

### Append range bug (`1eac654`)
The gws `+append` helper has no `--range` flag and always writes to `Sheet1`. Switched `append` off the helper onto `spreadsheets.values.append` (custom handler) so callers can target a specific tab. Default range remains `Sheet1` for backward compatibility. Extracted a shared `parseValuesInput` helper since `updateValues` and `append` take the same input shape.

### Queue enum gap (`cfa6c1d`)
The `queue_operations` schema enum listed only the four pre-factory tools (email/calendar/drive/accounts + scratchpad/workspace) — so agents couldn't queue operations against the factory-generated tools even though dispatch already handled them. Added `manage_sheets`, `manage_docs`, `manage_tasks`, `manage_meet` to the enum.

## Test plan

- [x] Unit tests — 26 new tests in `src/__tests__/factory/sheets-patch.test.ts` covering all formatters, all custom handlers, body shapes, param validation, and error paths
- [x] `patch-coverage.test.ts` snapshot — sheets now reports 13/13 custom formatted
- [x] `make check` — full CI gate green (383 tests)
- [x] `make coverage` — sheets gws coverage 32% → 47%
- [x] Live verified against a real Google account, each operation individually and then all together via `queue_operations`:
  - `create`, `get`, `read`, `getValues`, `updateValues`, `append`, `clearValues`
  - `addSheet`, `renameSheet`, `deleteSheet`, `duplicateSheet`, `renameSpreadsheet`, `copySheetTo`
  - Formula evaluation under `USER_ENTERED` (cascade through dependent cells confirmed)
  - `queue_operations` with 10 chained sheets operations on factory-generated tools
  - Bonus live test: built a working D&D 5e character sheet (Half-Elf Bard, L6) with reactive formulas — every save/skill/weapon attack/spell DC derived from ability scores and level

## Out of scope

- `manifest.yaml` crossed 1300 lines in this change — the quality hook flagged it. Worth a follow-up ADR to split per-service YAML files with the loader concatenating them.
- Remaining sheets gaps (10 operations): `*ByDataFilter` variants, multi-range batch reads/writes, `developerMetadata.*`. These are esoteric enough to defer.

Closes #90.

### DIFF
--- a/src/__tests__/factory/patch-coverage.test.ts
+++ b/src/__tests__/factory/patch-coverage.test.ts
@@ -148,6 +148,20 @@ describe('patch coverage', () => {
     expect(driveCustom).toContain('download');
   });
 
+  it('sheets core operations have custom formatting', () => {
+    const sheetsCustom = coverage
+      .filter(e => e.service === 'sheets' && e.usesCustomFormat)
+      .map(e => e.operation)
+      .sort();
+
+    expect(sheetsCustom).toContain('get');
+    expect(sheetsCustom).toContain('read');
+    expect(sheetsCustom).toContain('getValues');
+    expect(sheetsCustom).toContain('create');
+    expect(sheetsCustom).toContain('append');
+    expect(sheetsCustom).toContain('updateValues');
+  });
+
   it('meet core operations have custom formatting', () => {
     const meetCustom = coverage
       .filter(e => e.service === 'meet' && e.usesCustomFormat)
--- a/src/__tests__/factory/patch-coverage.test.ts
+++ b/src/__tests__/factory/patch-coverage.test.ts
@@ -154,12 +154,20 @@ describe('patch coverage', () => {
       .map(e => e.operation)
       .sort();
 
+    // Data ops
     expect(sheetsCustom).toContain('get');
     expect(sheetsCustom).toContain('read');
     expect(sheetsCustom).toContain('getValues');
     expect(sheetsCustom).toContain('create');
     expect(sheetsCustom).toContain('append');
     expect(sheetsCustom).toContain('updateValues');
+    // Tab management
+    expect(sheetsCustom).toContain('addSheet');
+    expect(sheetsCustom).toContain('renameSheet');
+    expect(sheetsCustom).toContain('deleteSheet');
+    expect(sheetsCustom).toContain('duplicateSheet');
+    expect(sheetsCustom).toContain('renameSpreadsheet');
+    expect(sheetsCustom).toContain('copySheetTo');
   });
 
   it('meet core operations have custom formatting', () => {
--- a/src/__tests__/factory/sheets-patch.test.ts
+++ b/src/__tests__/factory/sheets-patch.test.ts
@@ -1,0 +1,226 @@
+/**
+ * Tests for the sheets service patch — formatters preserve the `values`
+ * and `sheets` arrays that the generic detail formatter drops, and
+ * `updateValues` writes a request body the manifest can't express.
+ */
+
+jest.mock('../../executor/gws.js');
+import { execute } from '../../executor/gws.js';
+import { sheetsPatch } from '../../services/sheets/patch.js';
+import type { PatchContext } from '../../factory/types.js';
+
+const mockExecute = execute as jest.MockedFunction<typeof execute>;
+
+const ctx = (operation: string): PatchContext => ({
+  operation,
+  params: {},
+  account: 'user@test.com',
+});
+
+describe('sheetsPatch formatDetail', () => {
+  it('read renders the values array as a markdown table', () => {
+    const data = {
+      range: "'Clean Logs'!A1:C2",
+      majorDimension: 'ROWS',
+      values: [
+        ['Date', 'Level', 'Message'],
+        ['2026-04-15', 'INFO', 'started'],
+      ],
+    };
+    const res = sheetsPatch.formatDetail!(data, ctx('read'));
+    expect(res.text).toContain("'Clean Logs'!A1:C2");
+    expect(res.text).toContain('Date | Level | Message');
+    expect(res.text).toContain('2026-04-15 | INFO | started');
+    expect(res.text).toContain('**Rows:** 2');
+    expect(res.text).toContain('**Columns:** 3');
+    expect(res.refs.values).toEqual(data.values);
+    expect(res.refs.rowCount).toBe(2);
+  });
+
+  it('getValues uses the same values renderer', () => {
+    const res = sheetsPatch.formatDetail!(
+      { range: 'Sheet1!A1:B1', majorDimension: 'ROWS', values: [['a', 'b']] },
+      ctx('getValues'),
+    );
+    expect(res.text).toContain('a | b');
+  });
+
+  it('read handles empty range gracefully', () => {
+    const res = sheetsPatch.formatDetail!(
+      { range: 'Sheet1!A1:A1', majorDimension: 'ROWS' },
+      ctx('read'),
+    );
+    expect(res.text).toContain('**Rows:** 0');
+    expect(res.text).toContain('_(empty range)_');
+  });
+
+  it('escapes pipe characters so they do not break the table', () => {
+    const res = sheetsPatch.formatDetail!(
+      { range: 'S!A1', majorDimension: 'ROWS', values: [['a|b', 'c']] },
+      ctx('read'),
+    );
+    expect(res.text).toContain('a\\|b | c');
+  });
+
+  it('get renders spreadsheet metadata including sheet tabs', () => {
+    const data = {
+      spreadsheetId: 'sheet123',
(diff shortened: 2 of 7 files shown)


aaronsb:
## Code Review — feat/sheets-expansion

**Assessment:** The PR is largely solid. Fix for #90 is clean, the patch follows existing conventions closely, and test coverage is genuinely thorough (32 tests pass, all custom handlers exercised including the `sheetId=0` edge case). Architectural fit with the factory pattern is correct. One real bug found (next-steps guidance silently dropped), plus a handful of minor polish items.

### Issue (blocking): `nextSteps()` never fires for custom handlers

**Location**: `src/services/sheets/patch.ts` (all 8 `customHandlers` return paths) vs `src/server/formatting/next-steps.ts:130-175`.

**Problem**: The factory generator (`src/factory/generator.ts:181-183`) short-circuits to the custom handler and returns its response directly — it only appends `nextSteps(...)` on the non-custom-handler path (lines 224-230). Every other patch that has custom handlers (`docs/patch.ts:41,78`, `drive/patch.ts:55,92,…`, `calendar/patch.ts:217,230`, `gmail/patch.ts:251,…`) calls `nextSteps('domain', 'operation', { email: account })` inline in its return.

`sheetsPatch.customHandlers` never does. That means 8 operations — `updateValues`, `append`, `addSheet`, `renameSheet`, `deleteSheet`, `duplicateSheet`, `renameSpreadsheet`, `copySheetTo` — ship the guidance entries you carefully added to `next-steps.ts` but never render them. `create`, `get`, `read`, `getValues`, `clearValues` go through the factory path and DO get next-steps, so the regression is patch-local and silent.

**Why it matters**: Agents depend on the next-steps footer to chain operations naturally. Half the new surface loses that. And since the factory only wires next-steps on the non-custom path, this is a class of bug that will recur every time a new custom handler is added without an inline `nextSteps(...)` call.

**Suggestion**: Append `+ nextSteps('sheets', '<operation>', { email: account })` to each `text:` in the 8 handler returns. Example for `updateValuesHandler`:
```ts
return {
  text: parts.join('') + nextSteps('sheets', 'updateValues', { email: account }),
  refs: { /* ... */ },
};
```
A structural alternative worth considering (follow-up, not this PR): move the next-steps append into the generator so it happens post-custom-handler too. The current pattern requires every handler author to remember it, and this PR is evidence the convention doesn't hold. That would be a small refactor to `generator.ts:181-183` — wrap the custom handler result and append next-steps before returning. Worth an ADR.

### Minor

**1. Inconsistent index validation** (`src/services/sheets/patch.ts:321` vs `:399`)
`addSheetHandler` checks `params.index !== undefined && params.index !== null && params.index !== ''`.
`duplicateSheetHandler` checks `params.index !== undefined && params.index !== ''` (no `null`).
Passing `{ index: null }` to `duplicateSheet` yields `Number(null) === 0`, which passes `Number.isInteger(0) && 0 >= 0`, inserting the duplicate at position 0. Almost certainly not what a caller meant. Harmonize to the three-way check.

**2. `formatDetail` default fallthrough** (`:478-480`)
The `default` case in `formatDetail` routes unknown operations to `formatSpreadsheetDetail`. Currently fine because only `get`/`read`/`getValues` are `type: detail`. But if someone adds a new detail operation without updating the switch, it will silently mis-format. Prefer either an explicit `return formatSpreadsheetDetail(data)` with a comment, or a `formatDefault`-style fallback that generates something sensible for arbitrary data. Low priority.

**3. Naming inconsistency** (cosmetic)
Handlers use the `Handler` suffix (`updateValuesHandler`, `addSheetHandler`) but formatters don't (`formatValuesDetail`, `formatAppendAction`). Both patterns are self-consistent within their category, which is reasonable. Only worth noting because you asked — no change needed.

**4. `requireSheetId` accepts boolean coercion** (`:286`)
`Number(true) === 1`, which would pass `Number.isInteger`. A caller sending `{ sheetId: true }` gets `sheetId: 1`. Edge case, API will likely 404, but `typeof raw !== 'number' && typeof raw !== 'string'` upfront would be stricter. Nitpick.

### What I checked and why it's sound

- **SOLID**: The single `sheets/patch.ts` file groups helpers → formatters → custom handlers → export in a clear layered structure. At 511 lines it's comparable to `meet/patch.ts` (523) and under the 800-line priority threshold. `parseValuesInput`, `parseCsvLine`, `renderValuesTable`, `escapeCell`, `runBatchUpdate`, `requireSheetId` are all genuinely shared and narrowly-scoped. No accumulating responsibilities I'd split.
- **DRY**: `parseValuesInput` extraction (shared by `updateValues` and `append`) and `runBatchUpdate` (shared by 5 tab ops) are good — without them every handler would duplicate 20+ lines of boilerplate.
- **Manifest alignment**: Every custom handler's expected params match the manifest declaration. `spreadsheetId`/`range`/`values`/`jsonValues`/`valueInputOption` for values ops; `sheetId`/`title`/`index` for tab ops; `destinationSpreadsheetId` for copy. Verified line-by-line against `manifest.yaml:768-989`.
- **queue_operations enum fix** (`src/server/tools.ts:119`): Checked `src/server/handler.ts:31-40` — `domainHandlers` already registers all factory-generated tools. The dispatch was never narrow; only the schema enum was. Broadening it is safe and the right fix.
- **Error handling**: `throw new Error(...)` matches what `validate.ts` and other patches do. Error messages are specific and actionable (e.g., `"sheetId is required (integer, from manage_sheets get)"` points agents at the discovery step).
- **Test coverage for 26 tests**: The tests that matter are covered — body shape, `sheetId=0`, missing params, CSV vs JSON, range defaulting, non-integer rejection. Two gaps I'd note but wouldn't block on:
  - `parseValuesInput` CSV edge cases aren't directly tested (`"a,b"` with embedded quotes, `"""escaped"""`). They're exercised through `updateValues`/`append` happy paths, but the CSV parser has 20+ lines of logic that could silently regress.
  - `copySheetTo` error paths (missing `destinationSpreadsheetId`, non-integer `sheetId`) aren't tested.

### Summary

Ship blocker: the next-steps regression. Once those 8 inline calls land, this is good to merge. The rest is polish.

---

aaronsb:
Addressed all review items in 8ee3c8a:

**Blocker — nextSteps dropped for custom handlers**
Added `handlerContext()` helper that mirrors the factory's string-param context map, and applied `nextSteps('sheets', op, handlerContext(params, account))` to every one of the 8 custom handlers. Regression tests added that assert the footer lands AND that `<spreadsheetId>` resolves from handler context.

**Minor items**
- `duplicateSheetHandler`: aligned `index` null-check with `addSheetHandler`
- `requireSheetId`: explicit boolean rejection (previously `true` → `1`)
- `formatDetail` default: replaced silent fallback to `formatSpreadsheetDetail` with a generic key/value renderer (same pattern `meet/patch.ts` uses for unknown ops)
- Added tests: CSV with quoted commas, `""` escape sequences, `copySheetTo` missing-destination + execution-error paths

390 tests pass (was 383). No behavior change to the factory-path ops — their next-steps already worked via the generator.

The architectural follow-up you flagged (move next-steps append into the generator so it wraps custom handlers too, preventing this class of bug) is worth a dedicated ADR. Not included here to keep this PR's scope tight.

Issue #110, opened 2026-04-18
Windows: multi-word field values silently truncated in create (shell quoting bug)


JeremyYowell:
Root cause: gws.js spawns with shell: IS_WINDOWS = true, which causes Node.js to join the args array into a single command string without quoting. "Chesterfield YMCA North Court" becomes four separate tokens and gws rejects the extra words as unknown positional arguments. Single-word values work fine.

Affected operations: calendar create (fields: summary, location, description, attendees)
 
Fix (one change in executor/gws.js ~line 98):
// Before:
const proc = spawn(gwsBinary, fullArgs, { shell: IS_WINDOWS, ... });
 
// After:
const spawnBin  = IS_WINDOWS ? 'cmd.exe' : gwsBinary;
const spawnArgs = IS_WINDOWS ? ['/c', gwsBinary, ...fullArgs] : fullArgs;
const proc = spawn(spawnBin, spawnArgs, { /* shell: false (default) */ });
Using cmd.exe directly (required to run .cmd files on Windows) with the args as an array lets Node.js handle per-argument quoting at the OS level (CreateProcess), avoiding the shell-join problem entirely.
 
Version: 2.5.2 | Platform: Windows 11 (win32)

Issue #111, opened 2026-04-18
Windows: delete, update, and quickAdd fail with "Invalid --params JSON: key must be a string"


JeremyYowell:
Root cause: Same shell: IS_WINDOWS = true bug as issue #1, but manifesting differently. These operations pass structured data via --params '{"calendarId":"...","eventId":"..."}'. When Node.js shell mode builds the command string, cmd.exe consumes the " characters as string delimiters — stripping them — so gws receives {calendarId:...,eventId:...} (invalid JSON) instead of the original quoted string.

Affected operations: calendar delete, calendar update, calendar quickAdd (all --params JSON payloads)

Fix: Same one-line change as Issue #110  — using cmd.exe with an args array instead of shell: true prevents cmd.exe from mangling the embedded quotes.

Version: 2.5.2 | Platform: Windows 11 (win32)

PR #112, opened 2026-04-18
fix(windows): use cmd.exe /c for .cmd binaries to fix argument quoting

JeremyYowell:
## Problem

On Windows, two categories of operations are broken due to how `gws` is spawned:

**1. Multi-word field values are silently truncated (`create` and any op that passes string flags)**

Any calendar event field containing a space — `summary`, `location`, `description`, `attendees` — causes an error like:

```
error: unexpected argument 'YMCA' found
Usage: gws +insert [OPTIONS] --summary <TEXT> --start <TIME> --end <TIME>
```

**2. Structured `--params` JSON is corrupted (`delete`, `update`, `quickAdd`)**

These operations pass data as `--params '{"calendarId":"...","eventId":"..."}'`. The embedded double-quotes get consumed by the shell, producing malformed JSON:

```
Invalid --params JSON: key must be a string at line 1 column 2
```

## Root cause

Both failures stem from the same issue in `executor/gws.ts`. When `gws` is distributed as a Windows `.cmd` batch file, it cannot be spawned directly — `cmd.exe` is required to interpret it. The previous code handled this with `shell: true` on Windows, but that causes Node.js to join the entire args array into a single command string **without quoting individual elements**. Spaces in values split into separate tokens; double-quotes in JSON get stripped.

## Fix

When the resolved binary is a `.cmd` file, invoke `cmd.exe` directly with `/c` and pass the original args array unchanged. Node.js then handles per-argument quoting at the `CreateProcess` level, so spaces and special characters in values are preserved correctly.

```ts
// Before (broken on Windows):
const proc = spawn(gwsBinary, fullArgs, { shell: IS_WINDOWS, ... });

// After:
const isCmd = IS_WINDOWS && gwsBinary.endsWith('.cmd');
const spawnBin  = isCmd ? 'cmd.exe' : gwsBinary;
const spawnArgs = isCmd ? ['/c', gwsBinary, ...fullArgs] : fullArgs;
const proc = spawn(spawnBin, spawnArgs, { env, cwd, stdio: ['ignore', 'pipe', 'pipe'] });
```

Additionally, `resolveGwsBinary` now tries both `gws.exe` and `gws.cmd` as candidates (in that order), making it forward-compatible with native `.exe` distributions while remaining backward-compatible with legacy `.cmd` npm packages.

## Testing

Verified on Windows 11 with `@aaronsb/google-workspace-mcp` v2.5.2:
- Calendar `create` with multi-word `summary` and `location` now succeeds
- Calendar `delete` and `quickAdd` now succeed (JSON payload preserved correctly)

Closes #[issue numbers for the two Windows quoting bugs]

### DIFF
--- a/src/executor/gws.ts
+++ b/src/executor/gws.ts
@@ -23,7 +23,9 @@ const STALL_TIMEOUT = 30_000;   // Kill if no stdout/stderr activity for 30s
                                 // (generous: gws may do OAuth refresh, API pagination, cold start)
 
 const IS_WINDOWS = process.platform === 'win32';
-const GWS_BINARY_NAME = IS_WINDOWS ? 'gws.exe' : 'gws';
+// Prefer the native .exe (direct spawn, no shell needed).
+// Fall back to the .cmd wrapper used by legacy npm distributions.
+const GWS_BINARY_NAMES = IS_WINDOWS ? ['gws.exe', 'gws.cmd'] : ['gws'];
 
 /**
  * Resolve gws binary location.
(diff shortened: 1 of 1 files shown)


aaronsb:
Thanks for the thorough writeup and clean fix, @JeremyYowell! The root-cause analysis (`shell: true` joining args without quoting → spaces split, JSON quotes stripped) made this easy to review, and the `cmd.exe /c` approach is exactly right. Tests all green locally on Linux, and your Windows 11 verification covers the platform-specific behavior we couldn't unit-test from CI. Merged to main — closes #110 and #111. Appreciate the contribution!

Issue #114, opened 2026-04-21
manage_drive.listPermissions returns "No files found" for valid fileId


DirectCash-Personal:
## Summary

`manage_drive.listPermissions` returns the error text `No files found` even when the provided `fileId` refers to an existing, owned file. `get` on the same file returns proper metadata and other ops (`addComment`, `copy`, `download`, `delete`) work normally.

## Version

`@aaronsb/google-workspace-mcp@2.6.1` invoked via MCP stdio (direct `node build/index.js`, not through a gateway) on Linux, Node 20.

## Repro

1. Upload a file: `manage_drive` with `{"operation":"upload","email":"<user>","filePath":"/tmp/test.txt","name":"test"}` — returns `File ID: <ID>`.
2. Immediately call: `manage_drive` with `{"operation":"listPermissions","email":"<user>","fileId":"<ID>"}`.
3. Observed: response text is `## No files found.` plus the usual session-context block. `isError` is false.
4. Expected: list of permissions on the file (at minimum the owner's own permission).

Sequence was tested on a file that was just uploaded in the same MCP session and is confirmed to exist (`get` on the same `fileId` returns the file metadata successfully).

## Notes

- The error message "No files found" is the same message returned by `search` when no results match — suggests possible code path confusion between `search` and `listPermissions` in the handler or CLI translation layer.
- `share` operation on the same file returns a different error (`The permission type field is required.`) — not a scope issue; the request reaches the API.
- Would help narrow down if you can confirm whether the `listPermissions` op maps to `gws drive permissions list --params '{"fileId":"..."}'` or a different invocation.

Happy to provide additional logs / test another version / submit a PR once the root cause is identified.

PR #115, opened 2026-04-21
fix(calendar,drive): correct attendees flag, update schema, and share type

DirectCash-Personal:
## Summary

Three related bugs were surfaced while exercising the MCP against a live Google Workspace account on 2026-04-20 (version `2.6.1` via stdio). All three break real user flows with clear error signatures and are fixed together because they share a root cause: the default manifest generator routes everything through the gws CLI `--params` flag, which works for simple GET queries but not for helper commands or REST calls that need separate query params and JSON bodies.

### Bug 1 — `calendar.create` rejects every guest list

**Symptom:** `error: unexpected argument '--attendees' found; tip: a similar argument exists: '--attendee'`

`src/factory/manifest.yaml` declares `calendar.operations.create.cli_args.attendees: "--attendees"`, but `gws calendar +insert` (the `@googleworkspace/cli` helper) only accepts `--attendee` (singular). The calendar custom handler and the legacy `server/handlers/calendar.ts` had the same typo.

**Fix:** use `--attendee` in all three places. The singular flag accepts a comma-separated value.

### Bug 2 — `calendar.update` silently ignores attendees and Meet changes

**Symptom:** update returns `Operation completed` with HTTP 200, but the event is unchanged.

Two issues stacked:
- The manifest never declared `attendees` or `meet` under `update.params`, so `buildResourceArgs` silently dropped them.
- Even if declared, the default generator emits everything through `--params`, but Google Calendar's `events.patch` expects path/query fields (`eventId`, `calendarId`, `conferenceDataVersion`) in `--params` and the change body (`summary`, `start`, `end`, `attendees`, `conferenceData`) in `--json`.

**Fix:** declare `attendees` and `meet` on `update` in the manifest, and add an `update` customHandler on `calendarPatch` that:
- Routes eventId/calendarId through `--params` and the patch body through `--json` (same pattern used by `docsPatch.insertText`, `drivePatch.addComment`, etc.)
- Converts the comma-separated attendees string into `[{email: ...}]` as the API expects.
- Attaches `conferenceData.createRequest` + `conferenceDataVersion=1` when `meet: true` (Google's required shape for adding a Meet link).
- Rejects updates that carry no actual field change (the `at least one field to change` error).

Docstring notes that `events.patch` REPLACES the attendees array wholesale (Google does not diff arrays) and that removing an existing Meet link via patch is not supported by the API.

### Bug 3 — `drive.share` always 400s on `type` required

**Symptom:** `The permission type field is required.`

`src/factory/manifest.yaml` didn't declare `type` under `drive.operations.share.params`, and the default generator forced the whole permission object through `--params`. Drive v3's `permissions.create` wants the permission body (type, role, emailAddress|domain) in the request body, not the query string.

**Fix:** add a `share` customHandler on `drivePatch` that builds the JSON body and keeps only query params (fileId, supportsAllDrives, sendNotificationEmail) in `--params`. Manifest now accepts `type` (user|group|domain|anyone) and `domain`, defaulting `type` to `user` to preserve the prior implicit behavior. As a bonus, user/group shares no longer spam notification emails by default — you opt in by setting `sendNotificationEmail` yourself.

## Reproduction

Issued via stdio RPC to the compiled MCP (2.6.1):

```jsonc
// bug 1 — errors at the CLI layer
{"name":"manage_calendar","arguments":{"operation":"create","email":"u@x","summary":"t","start":"2026-04-22T10:00:00Z","end":"2026-04-22T11:00:00Z","attendees":"a@b.com"}}

// bug 2 — returns "Operation completed" but attendees remain empty
{"name":"manage_calendar","arguments":{"operation":"update","email":"u@x","eventId":"…","attendees":"a@b.com"}}

// bug 3 — API 400: "The permission type field is required."
{"name":"manage_drive","arguments":{"operation":"share","email":"u@x","fileId":"…","shareEmail":"a@b.com","role":"reader"}}
```

## Test plan

- [x] `npm run type-check`
- [x] `npm run build`
- [x] `npm test` — 33 suites / 596 passing (+17 new), 0 regressions
- [x] Updated the one existing test that asserted the broken `--attendees` value
- [x] New tests cover every branch of the two new custom handlers (update: no-fields error, summary-only, start/end dateTime mapping, attendees conversion, attendees clearing, Meet attach, explicit calendarId, refs.changed; share: body payload, type default, missing shareEmail, domain variant, anyone variant, notification suppression) and the `create` flag fix (asserts `--attendee`, rejects `--attendees`)
- [ ] (Recommended for maintainer) Manual smoke in a local Docker build against a real Google account to confirm the CLI accepts the new flags end-to-end

## Notes for review

- I kept the manifest `resource: events.patch` / `resource: permissions.create` entries in place so the factory still generates the correct schema/enum for the tool — the customHandler short-circuits before `buildResourceArgs` runs, so the manifest `defaults` are effectively inert for these two operations. Happy to strip the unused `resource` key if you'd prefer, but preserving it keeps the manifest self-documenting.
- There's a separate reported issue (`drive.listPermissions` returning "No files found" even on owned files) that I did not touch here — it needs a different investigation and I didn't want to scope-creep this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

**Related:** #114 (separate issue — `manage_drive.listPermissions` silent failure needing root-cause investigation before fix)

---

**Bug #2 scope clarification (added 2026-04-20 23:20 BRT):** empirical testing post-PR confirmed the bug scope is much broader than "attendees/meet" — it affects **every body field** in the update operation (summary, start, end, description, location, attendees, meet). Root cause: the generator pushes all params via `--params` (URL/query), but the `gws` CLI requires body fields to go via `--json`. The `customHandler` in this PR already separates them correctly, so the fix covers all fields; this note just corrects the commit message scope.

### DIFF
--- a/src/__tests__/factory/calendar-patch.test.ts
+++ b/src/__tests__/factory/calendar-patch.test.ts
@@ -282,5 +282,133 @@ describe('calendarPatch', () => {
 
       expect(result.refs.calendarId).toBe('primary');
     });
+
+    it('passes attendees via --attendee (singular) — gws rejects the plural form', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse(calendarInsertResponse));
+      await calendarPatch.customHandlers!.create(
+        { summary: 'X', start: 'Y', end: 'Z', attendees: 'a@b.com,c@d.com' },
+        'user@test.com',
+      );
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args).toContain('--attendee');
+      expect(args).not.toContain('--attendees');
+      const idx = args.indexOf('--attendee');
+      expect(args[idx + 1]).toBe('a@b.com,c@d.com');
+    });
+  });
+
+  describe('update custom handler', () => {
+    it('requires eventId', async () => {
+      await expect(
+        calendarPatch.customHandlers!.update({ summary: 'X' }, 'user@test.com'),
+      ).rejects.toThrow('eventId');
+    });
+
+    it('rejects updates with no fields to change', async () => {
+      await expect(
+        calendarPatch.customHandlers!.update({ eventId: 'evt-1' }, 'user@test.com'),
+      ).rejects.toThrow('at least one field');
+    });
+
+    it('routes via events.patch with --params and --json', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse({ id: 'evt-1', summary: 'New title' }));
+      await calendarPatch.customHandlers!.update(
+        { eventId: 'evt-1', summary: 'New title' },
+        'user@test.com',
+      );
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args.slice(0, 3)).toEqual(['calendar', 'events', 'patch']);
+      expect(args).toContain('--params');
+      expect(args).toContain('--json');
+
+      const queryParams = JSON.parse(args[args.indexOf('--params') + 1]);
+      expect(queryParams).toEqual({ calendarId: 'primary', eventId: 'evt-1' });
+
+      const body = JSON.parse(args[args.indexOf('--json') + 1]);
+      expect(body).toEqual({ summary: 'New title' });
+    });
+
+    it('maps start and end to dateTime objects', async () => {
+      mockExecute.mockResolvedValue(mockGwsResponse({ id: 'evt-1' }));
+      await calendarPatch.customHandlers!.update(
+        { eventId: 'evt-1', start: '2026-05-01T10:00:00Z', end: '2026-05-01T11:00:00Z' },
+        'user@test.com',
+      );
+
+      const args = mockExecute.mock.calls[0][0];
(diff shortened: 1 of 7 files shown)


aaronsb:
Thank you for this thorough fix, @DirectCash-Personal! The three-bug analysis (CLI flag mismatch on `calendar.create`, body-field routing on `calendar.update`, permission type on `drive.share`) and the 17 new tests made this a clear win to ship.

We landed your work via a maintainer-owned integration branch in #118 so we could (a) bundle in a one-line follow-up that picked up `src/server/scratchpad/adapters/send-calendar.ts` — a fourth call site that also had `--attendees` and that this PR didn't touch (caught by @adwitiyasingh11 in #117) — and (b) run CI on the maintainer-side before merging to main. Live verification against a real Google account confirmed all three fixes work end-to-end.

Your commit `123258a` is preserved in main's history via the merge — GitHub correctly auto-flagged this PR as merged. Really appreciate the contribution!

PR #119, opened 2026-05-11
fix(drive): listPermissions returns the permission list, not "No files found" (#114)

### LINKED ISSUE #114: manage_drive.listPermissions returns "No files found" for valid fileId

aaronsb:
## Problem

`manage_drive` `listPermissions` returned the error text `No files found` for any valid `fileId`, even one just uploaded in the same session. Reported in #114.

## Root cause

`listPermissions` is declared `type: list` in the manifest, and the drive patch's `formatList` is `formatFileList`, which reads `data.files`. But `permissions.list` returns `data.permissions` — so `formatFileList` saw an empty `files` array and emitted its "No files found." message. (Same structural family as the `share` body-vs-query bug already fixed: an operation whose response/request shape the generic path can't express.)

## Fix

Add a dedicated `listPermissions` custom handler (mirrors `share`):

- Requests an explicit `fields` mask — the Drive API's default `permissions.list` response omits `role`, `emailAddress`, and `domain`, which would have made the listing useless even once it rendered.
- Renders `permId | role | type | target` rows.
- Reports an empty permission list plainly instead of claiming "No files found".
- Labels `anyone`-with-link, pending-owner, and deleted-account permissions.

The manifest entry is unchanged (`type: list`, `resource: permissions.list`) — the custom handler takes precedence in `generateHandler`, so the generic `formatList` is no longer reached for this op.

## Verification

- `make build` + `make test` — all 609 tests pass (3 new for this handler).
- Live: `listPermissions` on an owned Google Doc now returns:
  ```
  ## Permissions on <id> (2)
  anyoneWithLink | writer | anyone | anyone with the link
  <id> | owner | user | owner@example.com
  ```

Closes #114.

### DIFF
--- a/src/__tests__/factory/drive-patch.test.ts
+++ b/src/__tests__/factory/drive-patch.test.ts
@@ -232,6 +232,69 @@ describe('drivePatch custom handlers', () => {
     });
   });
 
+  describe('listPermissions', () => {
+    it('hits permissions.list and requests the fields the API omits by default', async () => {
+      mockExecute.mockResolvedValueOnce({
+        success: true,
+        data: {
+          permissions: [
+            { id: 'owner-perm', type: 'user', role: 'owner', emailAddress: 'me@test.com' },
+            { id: 'reader-perm', type: 'user', role: 'reader', emailAddress: 'bob@test.com' },
+          ],
+        },
+        stderr: '',
+      });
+
+      const handler = drivePatch.customHandlers!.listPermissions!;
+      const result = await handler({ fileId: 'file-1' }, 'me@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args.slice(0, 3)).toEqual(['drive', 'permissions', 'list']);
+      const queryParams = JSON.parse(args[args.indexOf('--params') + 1]);
+      expect(queryParams.fileId).toBe('file-1');
+      // Without an explicit fields mask the Drive API returns only id/type — not
+      // role or emailAddress — which made the listing useless.
+      expect(queryParams.fields).toContain('role');
+      expect(queryParams.fields).toContain('emailAddress');
+
+      expect(result.text).toContain('Permissions on file-1 (2)');
+      expect(result.text).toContain('owner-perm | owner | user | me@test.com');
+      expect(result.text).toContain('reader-perm | reader | user | bob@test.com');
+      expect(result.refs.count).toBe(2);
+      expect(result.refs.permissionId).toBe('owner-perm');
+    });
+
+    it('reports an empty permission list without claiming "No files found"', async () => {
+      mockExecute.mockResolvedValueOnce({ success: true, data: { permissions: [] }, stderr: '' });
+
+      const handler = drivePatch.customHandlers!.listPermissions!;
+      const result = await handler({ fileId: 'file-2' }, 'me@test.com');
+
+      expect(result.text).toContain('No sharing permissions on file file-2');
+      expect(result.text).not.toContain('No files found');
+      expect(result.refs.count).toBe(0);
+    });
+
+    it("labels anyone-with-link and pending-owner permissions", async () => {
+      mockExecute.mockResolvedValueOnce({
+        success: true,
+        data: {
+          permissions: [
+            { id: 'anyone-perm', type: 'anyone', role: 'reader' },
+            { id: 'pending-perm', type: 'user', role: 'writer', emailAddress: 'new@test.com', pendingOwner: true },
+          ],
+        },
+        stderr: '',
+      });
+
+      const handler = drivePatch.customHandlers!.listPermissions!;
+      const result = await handler({ fileId: 'file-3' }, 'me@test.com');
+
+      expect(result.text).toContain('anyone-perm | reader | anyone | anyone with the link');
+      expect(result.text).toContain('[pending owner]');
+    });
+  });
+
   describe('download', () => {
     it('creates parent directories before calling gws', async () => {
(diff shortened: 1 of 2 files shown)


aaronsb:
## Code Review — fix(drive): listPermissions

**What this changes**: Adds a dedicated `listPermissions` custom handler in `src/services/drive/patch.ts` so the op no longer falls through `formatFileList` (which reads `data.files` and emitted "No files found" because `permissions.list` returns `data.permissions`). The handler also requests an explicit `fields` mask — the Drive API's default `permissions.list` response omits `role`/`emailAddress`/`domain` — and renders `id | role | type | target` rows with `[pending owner]` / `[deleted account]` labels.

**Assessment**: Correct fix, right architecture. Verified: `customHandlers[operation]` takes precedence over `formatList` in `generateHandler` (`src/factory/generator.ts:196`), so the manifest entry can stay `type: list` / `resource: permissions.list` unchanged. The choice of a custom handler over patching `formatFileList` is the correct call — the fix needs a non-default `fields` mask, which is request-shaping, not post-execute formatting. `formatFileList` couldn't reach it. Consistent with the `share` handler: query-only via `--params` (this op is a GET, so no `--json` body — easy place to get wrong, done right), comment block explains the *why* (the bug), and `execute()` rejects on non-zero exit so no `result.success` check is needed (matches every other handler in the file). 14/14 tests in `drive-patch.test.ts` pass.

### Findings

**[Medium] `refs.permissionId` and `refs.permissions` are misleading for a list result** — `src/services/drive/patch.ts:388-389`
For `share`, `permissionId` is well-defined (the perm just created). For a *list*, `permissions[0]?.id` is "whichever permission Drive returned first in API order" — frequently the caller themselves if they're the owner. The manifest's `unshare` param description literally says *"Permission ID to remove (from listPermissions)"*, inviting a `$N.permissionId` queue chain — which would silently target the wrong row. And `refs.permissions` (the array) is also a footgun: `resolveRef` in `src/server/queue.ts:165` does `String(extracted)` unconditionally, so `$N.permissions` stringifies to `owner-perm,reader-perm` — inert as a single arg.
Suggestion: drop `permissionId` from the list refs (or rename to something that signals arbitrariness, e.g. `firstPermissionId`); keep `permissions: [...]` (it's the honest answer for a human reader / structured consumer) but be aware the queue placeholder can't use it meaningfully.

**[Low] No pagination** — declared `type: list` but the `fields` mask omits `nextPageToken` and the handler accepts no `pageSize`/`pageToken`. Permission lists on heavily-shared files (or shared drives) can exceed one page. Not a regression — `listComments` in the same file has the identical shape — so this is a known codebase gap, not a blocker on this PR. Worth a follow-up issue if shared-drive files are in scope.

**[Nit] Render branches untested** — the `domain` target, `displayName` fallback, and `deleted`-flag rendering paths aren't exercised by the 3 new tests (the `pendingOwner` + `anyone` paths are). Three more lines of mock data would close the gap. Optional.

### Acknowledgements
- Comment in the handler explains the bug, mirroring `share`'s comment block — good.
- Empty-list message (`No sharing permissions on file <id>.`) is plain and correct — and the test explicitly asserts it does *not* contain "No files found", which directly pins the regression.
- `--params` carries query-only; correctly does not add a `--json` body for this GET.

### Verdict
Not blocking. The `refs.permissionId` item is worth addressing before merge (one-line change) since the manifest actively suggests the chaining pattern it breaks; the rest are optional/follow-up. Ship-ready otherwise.

---
*AI-assisted review via Claude Code (code-reviewer)*

aaronsb:
Addressed the code review in c499bc8:
- Dropped the misleading singular `refs.permissionId` (it was `permissions[0].id` — an arbitrary row, usually the owner, which the `unshare` param description would mislead callers into reusing).
- `refs.permissions` is now an array of `{permissionId, type, role, target}` objects so a caller can pick the specific row to unshare.
- Added render-branch coverage for domain / displayName / deleted-account targets.

The pagination + manifest-shape gap (no `nextPageToken`/`pageSize`) is left as-is — it's a pre-existing codebase pattern shared with `listComments`; worth a separate follow-up issue if shared-drive scaling matters.

PR #120, opened 2026-05-11
fix(drive): copy honors `name`; add `update` op for rename/move (#105)

### LINKED ISSUE #105: drive: `copy` drops `name`; rename/move absent in post-refactor patch layer
Two related gaps in the Drive service after the toolkit refactor that closed #48. Both map cleanly to the factory + service-patch architecture.

### 1. `manage_drive` `copy` silently ignores `name`

**Repro**
```json
{ "operation": "copy", "email": "…", "fileId": "<src>", "name": "My New Name" }
```
**Expected:** new file titled `My New Name`.
**Actual:** new file titled `Copy of <original>`. The `name` parameter is accepted (no validation error) but has no effect.

**Why it happens**

`src/services/drive/patch.ts` has a dedicated `upload` custom handler that explicitly forwards `params.name` to the underlying CLI:

```ts
if (params.name) args.push('--name', String(params.name));
```

There is no `copy` custom handler, so the call flows through the generic manifest-driven path, and the manifest entry for the `copy` operation doesn't map `name` onto the Drive `files.copy` request body. The Drive API supports `name` on the initial `files().copy()` call, so this is a one-request fix — no follow-up `update`.

**Fix options** (in increasing surface area):
1. Extend the generic handler so declared params flow into the copy request body when the operation supports it (preferred if other services will hit the same pattern).
2. Add a `copy` entry to `drivePatch.customHandlers` mirroring the `upload` pattern: forward `name` and `parentFolderId` to the gws `drive files copy` invocation.
3. At minimum: remove `name` from the `copy` branch of the schema so callers aren't misled.

### 2. No `rename` / `move` / `update` operation in Drive patch

Issue #48 was closed as *"superseded by the complete toolkit refactor,"* but the refactored `drivePatch.customHandlers` and the `drive` operation enum in `manage_drive` still expose no way to:

- Rename an existing Drive file in place
- Move a file between folders
- Update file metadata generally

Current workaround — download → re-upload with new name → delete original — round-trips native Google Sheets/Docs through xlsx/docx and risks formatting drift on non-trivial files.

**Suggested addition** to `drivePatch.customHandlers`:

```ts
update: async (params, account): Promise<HandlerResponse> => {
  const fileId = requireString(params, 'fileId');
  const body: Record<string, unknown> = {};
  if (params.name) body.name = params.name;
  const args = ['drive', 'files', 'update',
    '--params', JSON.stringify({ fileId, supportsAllDrives: true, body })];
  if (params.parentFolderId) { args.push('--add-parents', String(params.parentFolderId)); }
  // …
}
```

…paired with an `update` entry in the `drive` section of `manifest.yaml` exposing `fileId`, `name`, and `parentFolderId` (and `removeParents` if move-between-folders is in scope). This gives one primitive for rename, move, or rename-and-move.

### Environment

- Hit while doing an invoice workflow: copied a Google Sheet, expected `name` to rename the copy; had to fall back to the Drive UI.
- Version: whatever MCP is serving `manage_drive` as of 2026-04-15.

aaronsb:
## Problem (#105)

1. `manage_drive` `copy` accepted a `name` param but silently dropped it — copies kept Drive's `"Copy of <original>"` default name.
2. After the toolkit refactor there was no way to rename a file in place or move it between folders.

## Root cause

Same body-vs-query split that `share` already hit: the generic factory path (`buildResourceArgs`) puts every declared param into the query `--params`, but Drive's `files.copy` and `files.update` take renamable metadata (`name`, `parents`) in the **request body**.

## Fix

- **`copy`** — new custom handler: `name` → `body.name`, `parentFolderId` → `body.parents`, sent via `--json`; requests `id, name, parents, webViewLink` back. Omits `--json` entirely when neither is given (plain copy still works).
- **`update`** — new operation (manifest entry + custom handler): `name` renames (request body); `addParents` / `removeParents` move between folders (query params — `files.update` keeps parent changes in the query, renamable metadata in the body). Rejects a no-op call where none of `name` / `addParents` / `removeParents` is set.

Both `copy` and the new `update` op are surfaced in `manage_drive`'s operation enum with descriptions that explain the rename/move semantics.

## Verification

- `make build` + `make test` — all 615 tests pass (6 new for `copy` / `update`).
- `make manifest-lint` — passes (`manage_drive` now 17 operations).
- Live, against a real Google account:
  - `copy` on a Google Doc with `name: "MCP copy/update repro — DELETE ME"` → new file titled exactly that (not "Copy of …").
  - `update` with `name: "… RENAMED …"` → file renamed.
  - test file deleted afterward.

Closes #105.

### DIFF
--- a/src/__tests__/factory/drive-patch.test.ts
+++ b/src/__tests__/factory/drive-patch.test.ts
@@ -307,6 +307,92 @@ describe('drivePatch custom handlers', () => {
     });
   });
 
+  describe('copy', () => {
+    it('sends name as the request body, not via --params, so the copy is renamed', async () => {
+      mockExecute.mockResolvedValueOnce({
+        success: true,
+        data: { id: 'copy-1', name: 'My New Name', mimeType: 'application/vnd.google-apps.spreadsheet' },
+        stderr: '',
+      });
+
+      const handler = drivePatch.customHandlers!.copy!;
+      const result = await handler({ fileId: 'src-1', name: 'My New Name' }, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args.slice(0, 3)).toEqual(['drive', 'files', 'copy']);
+      const body = JSON.parse(args[args.indexOf('--json') + 1]);
+      expect(body.name).toBe('My New Name');
+      const queryParams = JSON.parse(args[args.indexOf('--params') + 1]);
+      expect(queryParams.fileId).toBe('src-1');
+      // name must NOT leak into the query — that's the bug being fixed.
+      expect(queryParams.name).toBeUndefined();
+
+      expect(result.text).toContain('File copied: **My New Name**');
+      expect(result.text).toContain('**File ID:** copy-1');
+      expect(result.refs.fileId).toBe('copy-1');
+      expect(result.refs.sourceFileId).toBe('src-1');
+    });
+
+    it('forwards parentFolderId as body.parents', async () => {
+      mockExecute.mockResolvedValueOnce({ success: true, data: { id: 'copy-2', name: 'Copy of X', parents: ['folder-9'] }, stderr: '' });
+
+      const handler = drivePatch.customHandlers!.copy!;
+      await handler({ fileId: 'src-2', parentFolderId: 'folder-9' }, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      const body = JSON.parse(args[args.indexOf('--json') + 1]);
+      expect(body.parents).toEqual(['folder-9']);
+    });
+
+    it('omits --json entirely when neither name nor parentFolderId is given', async () => {
+      mockExecute.mockResolvedValueOnce({ success: true, data: { id: 'copy-3', name: 'Copy of X' }, stderr: '' });
+
+      const handler = drivePatch.customHandlers!.copy!;
+      await handler({ fileId: 'src-3' }, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args).not.toContain('--json');
+    });
+  });
+
+  describe('update', () => {
+    it('renames via the request body and surfaces the new name', async () => {
+      mockExecute.mockResolvedValueOnce({ success: true, data: { id: 'file-1', name: 'Renamed.pdf', parents: ['root'] }, stderr: '' });
+
+      const handler = drivePatch.customHandlers!.update!;
+      const result = await handler({ fileId: 'file-1', name: 'Renamed.pdf' }, 'user@test.com');
+
+      const args = mockExecute.mock.calls[0][0];
+      expect(args.slice(0, 3)).toEqual(['drive', 'files', 'update']);
+      const body = JSON.parse(args[args.indexOf('--json') + 1]);
+      expect(body.name).toBe('Renamed.pdf');
+      const queryParams = JSON.parse(args[args.indexOf('--params') + 1]);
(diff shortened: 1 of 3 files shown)


aaronsb:
## Review — fix(drive): copy honors \`name\`; add \`update\` op

**Assessment:** Solid. The body-vs-query split is correct per the Drive v3 API (\`files.copy\` takes \`name\`/\`parents\` in the request body; \`files.update\` keeps \`addParents\`/\`removeParents\` as query params and renamable metadata in the body), it cleanly extends the established \`share\`/\`upload\`/\`download\` custom-handler pattern in the same file, framing/next-steps still flow through the generator per ADR-303, and both ops were verified live. No blocking issues.

### Findings

**[low] Test gap — combined rename+move on \`update\`**
\`src/__tests__/factory/drive-patch.test.ts\` covers pure-rename (body only) and pure-move (query only) for \`update\`, but not the cross-cutting case where \`name\` + \`addParents\` + \`removeParents\` are all set on one call — i.e. \`--json\` and \`--params\` both populated in the same request. That's the most interesting branch of the handler; worth one more test asserting \`body.name\` and \`queryParams.addParents\`/\`removeParents\` coexist.

**[low] \`parentFolderId\` (copy) vs \`addParents\`/\`removeParents\` (update) naming — reasonable, accept**
The asymmetry is justified: a \`update\` *move* needs the dual remove-old + add-new semantic, which a single \`parentFolderId\` can't express, and matching the Drive API's own field names is clearer than inventing wrappers. \`copy\`'s single-destination \`parentFolderId\` correctly mirrors \`upload\`. No change needed — flagging only because the task asked.

**[nit] \`refs.parents\` can be \`undefined\` in \`update\`**
\`src/services/drive/patch.ts\` — when the API response has no \`parents\` array, \`refs: { ..., parents }\` carries an explicit \`undefined\`. Harmless under \`JSON.stringify\`, but slightly inconsistent with how the rest of the file conditionally spreads optional refs. Take it or leave it.

**[observation] Inert manifest metadata on custom-handler ops**
\`copy\` and \`update\` declare \`resource: files.copy\`/\`files.update\` and \`defaults: supportsAllDrives: true\`, but the custom handlers hardcode the resource path and \`supportsAllDrives\` in their own \`--params\` JSON, so that manifest metadata never executes (the generic \`buildResourceArgs\` path is shadowed by \`customHandlers[operation]\`). This is the same shape as the existing \`share\` entry, so it's a pre-existing pattern, not a defect in this PR — noting it so a future manifest edit doesn't expect the handler to pick up a changed default.

### Considerations / future work
- \`copy\` and \`update\` have no \`next-steps\` entries — \`nextSteps()\` returns empty gracefully, and \`export\` has the same gap, so this is consistent. Could add (e.g. copy → \`get\` the new file; update → \`get\` to confirm) when convenient.
- The \`update\` description's "to move, set both" is a fair hint that setting only \`addParents\` won't behave like a move on modern single-parent Drive. Adequate as worded; mentioning a follow-up only if user feedback shows confusion.

---
*AI-assisted review via Claude Code*

aaronsb:
Addressed the review in 56c54cb:
- `update` now spreads `parents` into `refs` only when the API returns them (was an explicit `undefined`).
- Added the combined rename+move test (the branch that populates both `--json` and `--params`) plus a "response has no parents" ref test. 22 tests in drive-patch.test.ts now.

Inert manifest metadata (`resource:` / `defaults:` on `copy`/`update`) left as-is — same pre-existing shape as the `share` entry; harmonizing that is a separate cleanup if wanted.

PR #121, opened 2026-05-11
fix(sheets): create honors `title`; read/getValues render row numbers (#113)

### LINKED ISSUE #113: sheets: `create` drops `title`; `read`/`getValues` rendering strips blank rows
Two findings from the `manage_sheets` service — both surfaced while working through the `manage_drive` issues in #105, then isolated in blank-sheet repros.

### 1. `manage_sheets` `create` silently ignores `title`

**Repro**
```json
{ "operation": "create", "email": "…", "title": "MCP updateValues repro — DELETE ME" }
```
**Expected:** new spreadsheet named `MCP updateValues repro — DELETE ME`.
**Actual:** new spreadsheet named `Untitled spreadsheet`. `title` is accepted without validation error but has no effect.

**Why it happens (same structural pattern as #105)**

`src/services/sheets/patch.ts` defines a `formatCreateAction` formatter but no `create` entry in `customHandlers`. So the call flows through the generic manifest-driven path, and the manifest's `sheets` `create` operation doesn't map `title` into the Drive `spreadsheets.create` request body (`properties.title`). The Sheets API supports setting title on the initial create call — no follow-up rename needed.

Workaround today: `create` → then `renameSpreadsheet` as a second call (which works — it already has a custom handler and was what I used to rename `Copy of JVL Invoice 1080 …` after `manage_drive.copy` ate `name`).

**Fix** mirrors the `copy`+`name` fix proposed in #105 — either extend the generic handler so declared params flow into the create request body, or add a `create` custom handler in `sheetsPatch` that sets `{ properties: { title } }` on the spreadsheet being created.

### 2. `read` / `getValues` rendering strips blank rows (at least in display)

During the Invoice 1081 workflow I wrote values via `updateValues` to `G2`, `F5`, `F15`, `H21`. They all appeared in read-backs to have landed one row too high (e.g. `4/20/2026` written to `G2` visually rendered above the row containing `Aaron Bockelie`). I then opened a bug suspecting an off-by-one in `updateValues`.

After isolating in a blank spreadsheet with known targets:

| Write target | Rendered position on readback |
|--------------|-------------------------------|
| `A1`         | A1 ✓                          |
| `B2`         | B2 ✓                          |
| `C3`         | C3 ✓                          |
| `D4`         | D4 ✓                          |

So `updateValues` is correct. What actually happened in the template sheet is that the rendered output of `read`/`getValues` appears to strip or collapse leading / fully-blank rows, making the rendered "row 1" line correspond to sheet row 2 (or wherever the first non-empty row actually is). Counting rows against the render led me to under-shoot by one on every write target.

A user can't see from the rendered output that a blank row exists — it just looks like row 1 starts at the first non-empty row. The write still works correctly, but the tool's own read output is misleading for authoring subsequent writes.

**Suggestions, roughly increasing effort:**
- Prefix each rendered row with its actual sheet row number (`R2: | Aaron Bockelie | …`) — makes the render directly usable for targeting follow-up writes.
- Or: emit placeholder lines for blank rows so line-counting is preserved (`R1: [empty]`).
- At minimum: document in the `read`/`getValues` tool description that the rendered output skips empty rows so authors don't count off it.

### Environment
- Hit while doing the same JVL invoice workflow as #105.
- Version: whatever MCP is serving `manage_sheets` as of 2026-04-20.

Related: #105 (drive: `copy` drops `name`; rename/move absent). Same "declared param silently dropped by generic manifest path" family for point 1.

aaronsb:
## Problem (#113)

1. `manage_sheets` `create` accepted a `title` param but silently dropped it — every new spreadsheet came out `"Untitled spreadsheet"`.
2. `read` / `getValues` rendered output stripped blank rows, so a reader counting lines off the render under-shot write targets by however many leading rows were blank (the symptom: `4/20/2026` written to `G2` "appeared" above the `Aaron Bockelie` row).

## Root cause

1. Same body-vs-query split as `share` / `drive copy`: `spreadsheets.create` takes the spreadsheet's `properties` (incl. `title`) in the request **body**, but the generic factory path collapses every declared param into the query `--params` JSON.
2. The Sheets API trims trailing-empty rows and may omit leading-empty rows from the `values` array; `renderValuesTable` emitted one line per array element with no row number, so line N of the render ≠ sheet row N.

## Fix

- **`create`** — new custom handler sends `{ properties: { title } }` via `--json` (omits `--json` entirely when no title is given). `title` is now a documented param on the op; description updated.
- **`read` / `getValues` render** — each rendered row is prefixed with its actual sheet row number (`R3: a | b | c`), derived from the response `range` (`Sheet1!A3:D10` → start row 3). A fully-blank row renders as a bare `Rn:` so it's visible. Numbers are right-aligned so the colons line up. `refs.startRow` added. Op descriptions mention the prefix.

Existing `formatAction`'s `create` branch is kept (it's now unreachable for live calls since the custom handler intercepts, but a unit test exercises it directly and it's a harmless fallback).

## Verification

- `make build` + `make test` — all 622 tests pass (12 new).
- `make manifest-lint` — passes (`manage_sheets` still 13 operations).
- Live, against a real Google account:
  - `create` with `title: "MCP #113 repro — DELETE ME"` → sheet named exactly that.
  - `read` of a sheet whose rows 1–2 are blank → renders `R1:` / `R2:` then `R3: Aaron Bockelie | Invoice | 1081` …
  - `read` of `Sheet1!A3:C5` → rows numbered `R3`–`R5`.
  - test sheet deleted afterward.

Closes #113.

### DIFF
--- a/src/__tests__/factory/sheets-patch.test.ts
+++ b/src/__tests__/factory/sheets-patch.test.ts
@@ -18,7 +18,7 @@ const ctx = (operation: string): PatchContext => ({
 });
 
 describe('sheetsPatch formatDetail', () => {
-  it('read renders the values array as a markdown table', () => {
+  it('read renders the values array with sheet row-number prefixes', () => {
     const data = {
       range: "'Clean Logs'!A1:C2",
       majorDimension: 'ROWS',
@@ -29,20 +29,58 @@ describe('sheetsPatch formatDetail', () => {
     };
     const res = sheetsPatch.formatDetail!(data, ctx('read'));
     expect(res.text).toContain("'Clean Logs'!A1:C2");
-    expect(res.text).toContain('Date | Level | Message');
-    expect(res.text).toContain('2026-04-15 | INFO | started');
+    expect(res.text).toContain('R1: Date | Level | Message');
+    expect(res.text).toContain('R2: 2026-04-15 | INFO | started');
     expect(res.text).toContain('**Rows:** 2');
     expect(res.text).toContain('**Columns:** 3');
     expect(res.refs.values).toEqual(data.values);
     expect(res.refs.rowCount).toBe(2);
+    expect(res.refs.startRow).toBe(1);
   });
 
-  it('getValues uses the same values renderer', () => {
+  it('numbers rows from the start of the requested range, not from 1', () => {
+    // The Sheets API omits leading-blank rows from `values`; the response
+    // `range` is what tells you where row 1 of the array actually sits.
+    const res = sheetsPatch.formatDetail!(
+      {
+        range: "'Invoice'!G2:H4",
+        majorDimension: 'ROWS',
+        values: [['4/20/2026', ''], ['Aaron Bockelie', ''], ['', '$1,234']],
+      },
+      ctx('read'),
+    );
+    expect(res.text).toContain('R2: 4/20/2026 |');
+    expect(res.text).toContain('R3: Aaron Bockelie |');
+    expect(res.text).toContain('R4:  | $1,234');
+    expect(res.refs.startRow).toBe(2);
+  });
+
+  it('renders a fully-blank row as a bare Rn: so it is visible', () => {
+    const res = sheetsPatch.formatDetail!(
+      { range: 'Sheet1!A1:B3', majorDimension: 'ROWS', values: [[], [], ['data', 'here']] },
+      ctx('getValues'),
+    );
+    expect(res.text).toContain('R1:\n');
+    expect(res.text).toContain('R2:\n');
+    expect(res.text).toContain('R3: data | here');
+  });
+
+  it('pads row numbers so the colons line up across digit widths', () => {
+    const values = Array.from({ length: 11 }, (_, i) => [`v${i + 1}`]);
+    const res = sheetsPatch.formatDetail!(
+      { range: 'Sheet1!A1:A11', majorDimension: 'ROWS', values },
+      ctx('read'),
+    );
+    expect(res.text).toContain('R 1: v1');
+    expect(res.text).toContain('R11: v11');
+  });
+
+  it('getValues uses the same row-prefixed renderer', () => {
     const res = sheetsPatch.formatDetail!(
       { range: 'Sheet1!A1:B1', majorDimension: 'ROWS', values: [['a', 'b']] },
       ctx('getValues'),
     );
-    expect(res.text).toContain('a | b');
+    expect(res.text).toContain('R1: a | b');
   });
 
   it('read handles empty range gracefully', () => {
@@ -59,7 +97,7 @@ describe('sheetsPatch formatDetail', () => {
       { range: 'S!A1', majorDimension: 'ROWS', values: [['a|b', 'c']] },
       ctx('read'),
     );
-    expect(res.text).toContain('a\\|b | c');
+    expect(res.text).toContain('R1: a\\|b | c');
   });
 
   it('get renders spreadsheet metadata including sheet tabs', () => {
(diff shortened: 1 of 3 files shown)


aaronsb:
## Review Summary

**What this changes**: Fixes both arms of #113 in `manage_sheets`. (1) `create` now sends `{ properties: { title } }` via `--json` through a new `createSpreadsheetHandler` (registered in `customHandlers`), matching the `share` / `drive copy` body-vs-query pattern; the `title` param is documented on the op. (2) `read` / `getValues` render now prefixes each row with its real sheet row number (`R3: a | b | c`), derived from the response `range` via a new `startRowFromRange` helper; fully-blank rows render as a bare `Rn:`; numbers are right-aligned. `refs.startRow` added; op descriptions updated; 12 new tests.

**Assessment**: Solid, well-scoped, follows established conventions. No blockers. The custom-handler intercept in `generator.ts` runs before the generic path, so the body-vs-query fix is effective and `formatAction`'s now-unreachable `create` branch is a harmless fallback (still exercised by a direct unit test of `formatCreateAction`). `execute(args, { account })` is consistent with the other sheets handlers (`format` defaults to `json`). `startRowFromRange` handles the cases that matter: `Sheet1!A3:D10` -> 3, `'My Sheet'!B5:D9` -> 5 (uses `lastIndexOf('!')`), `Sheet1` with no `!` -> 1, column-only `Sheet1!A:D` -> 1 (correct, since that range does start at row 1). Test traces check out, including the two-space `R4:  | $1,234` case (empty first cell + ` | ` join) and the padding cases (`R 1:` / `R11:`). Backward-compat holds: existing `.toContain('a | b')`-style assertions still pass because `R1: a | b` contains the substring.

### Findings (all Low severity / non-blocking)

1. **`title` param description regression at the tool-schema level.** `generateSchema` collapses params across operations and keeps the *first* declaration's description. Before this PR the first `title:` in the sheets block was `addSheet`'s ("Name of the new sheet tab"); now `create` is declared earlier in the operations list, so the consolidated schema-level `title` description flips to "Spreadsheet title (defaults to 'Untitled spreadsheet')" — slightly misleading when the model is choosing `title` for `addSheet` / `renameSheet` / `duplicateSheet` / `renameSpreadsheet`. The per-operation descriptions in the `operation` enum text remain authoritative, so this is cosmetic, but worth a decision: (a) make `create`'s `title` description more neutral ("title for the new spreadsheet, or the sheet tab name for tab operations"), (b) rename the param on `create` (e.g. `spreadsheetTitle`), or (c) accept as-is.

2. **`majorDimension === 'COLUMNS'` would mislabel the prefix.** `formatValuesDetail` always passes `startRowFromRange(range)` regardless of `majorDimension`; if a response ever came back as `COLUMNS`, `values[0]` is the first *column*, not row 1, and `Rn:` would be wrong. This is unreachable today — neither `read` (the `+read` helper) nor `getValues` exposes a `majorDimension` param, so callers can't request it and the API default is `ROWS`. Safe now, fragile to a future param addition. Suggest either gating the prefix (`if (majorDimension === 'ROWS')`, fall back to the un-prefixed render otherwise) or a one-line comment in `renderValuesTable` / `formatValuesDetail` documenting the assumption.

3. **Keeping the dead `formatAction` `case 'create'`** is the right call — it makes `formatCreateAction` a guarded primitive even if dispatch ever flips back. A one-line comment (`// fallback — live calls intercept via customHandlers.create`) would make that intent explicit in the code, not just the PR body.

4. **Render is not round-trip-parseable** — a cell whose first value happens to start with `R<digits>:` collides visually with the row prefix. Not a correctness bug (this is a human/agent-facing render, not a parse target), just worth being aware of. No action needed.

5. **`startRowFromRange` with a tab name containing `!`** would land `lastIndexOf('!')` inside the name. Theoretical — Sheets disallows `!` in tab names in practice, and the prior code had the same exposure. No action needed.

### Considerations

- Worth a quick manifest-lint / build confirmation in CI (PR body says 622 tests pass, 12 new, `make manifest-lint` clean, plus live verification against a real account — looks thorough).
- `next-steps.ts` already has a `sheets.create` entry, so the framing footer after the new handler is fine without changes.

aaronsb:
Addressed the review in 9dadf5c (both non-blocking nits → comments, no behavior change):
- documented in `formatValuesDetail` why the `Rn:` prefix is always safe (`read`/`getValues` never expose `majorDimension` → always ROWS-major; a future COLUMNS path would need `Cn:`).
- marked `formatAction`'s `case 'create'` as the unreachable-for-live fallback that `customHandlers.create` shadows (kept because a direct unit test exercises `formatCreateAction` through it).

Finding #1 (consolidated tool-schema `title` description now reads `create`'s wording, since `create` is declared before the tab ops): leaving as-is — the per-op enum descriptions are authoritative, and this matches the existing pattern for `range` (described from `read`'s perspective but shared by `append`/`getValues`/etc.).

PR #122, opened 2026-05-11
feat(manifest): split manifest.yaml into per-service files (ADR-304, #108)

### LINKED ISSUE #108: Implement ADR-304: Split manifest.yaml into per-service files
Implementation tracking for [ADR-304](../blob/main/docs/architecture/api/ADR-304-split-manifest-yaml-into-per-service-files.md), accepted but deferred.

## Motivation

`src/factory/manifest.yaml` is 1383 lines and has been flagged by the quality-check hook on consecutive PRs (#103, #107). Each service's block is self-contained — the file is a concatenation of independent sections that happen to share a YAML root.

## What needs to happen

Per ADR-304:

1. Create `src/factory/manifest/` with one file per service (`gmail.yaml`, `calendar.yaml`, `drive.yaml`, `sheets.yaml`, `docs.yaml`, `tasks.yaml`, `meet.yaml`). Each file's root is the service definition — no enclosing `services:` key. Filename = service key.
2. Update `loadManifest()` in `src/factory/generator.ts` to enumerate the directory and assemble the `Manifest`.
3. Update `Makefile`:
   - `build:` — copy `src/factory/manifest/` recursively instead of the single file
   - `manifest-lint:` — walk the directory
   - `mcpb:` — ship the directory in the bundle
4. Delete `src/factory/manifest.yaml`.

The split is mechanical — cut each section into its own file, adjust indentation up one level.

## Acceptance criteria

- [ ] No YAML file above ~300 lines
- [ ] `make check` passes with no test changes needed (`Manifest` type and `generateTools()` API unchanged)
- [ ] `make coverage` output unchanged
- [ ] mcpb bundle still loads the manifest correctly

## Scope

Out of scope for this issue: co-locating manifest files with patches under `src/services/<name>/` (considered as an alternative in the ADR — could be a follow-up).

aaronsb:
Implements [ADR-304](https://github.com/aaronsb/google-workspace-mcp/blob/main/docs/architecture/api/ADR-304-split-manifest-yaml-into-per-service-files.md). Closes #108.

## What changed

`src/factory/manifest.yaml` (~1420 lines, flagged by the quality-check hook on consecutive PRs) → `src/factory/manifest/`, one file per service:

```
src/factory/manifest/
  gmail.yaml      calendar.yaml   drive.yaml      sheets.yaml
  docs.yaml       tasks.yaml      meet.yaml       README.md
```

Each file's YAML root *is* that service's definition (no enclosing `services:` key); the filename minus `.yaml` is the service key.

- **`loadManifest()`** (`src/factory/generator.ts`) now enumerates the manifest directory and assembles a `Manifest` with the same shape it returned before. The module-relative → cwd search strategy is reused, pointed at the directory. A malformed file fails the whole load (`parseYaml` throws) — same whole-or-nothing behavior as the single file. The optional arg is now a directory path.
- **Makefile**: `build` copies the directory to `build/factory/manifest/`; `manifest-lint` walks it; `manifest-diff` reassembles a `services:`-wrapped temp file to diff against `discovered-manifest.yaml`; `mcpb` ships the directory. `package.json`'s `build` script kept in sync.
- File-level notes (contacts exclusion, Meet scopes) moved to `src/factory/manifest/README.md`. Factory way doc, README diagram, and `gen-manifest.sh` comments updated to the new layout.

## How the split was verified

Script-generated, then round-tripped: parse the original `manifest.yaml`, parse each new per-service file, reassemble into `{services: {...}}`, and assert the two parse-trees are deep-equal. The reassembled-from-directory manifest also matches `loadManifest('src/factory/manifest')` exactly.

The `Manifest` type and `generateTools()` API are unchanged, so **no test changes** — all 622 tests pass. `make manifest-lint` reports the same 80 operations across 7 services. The built registry loads 7 services / 7 tools from the directory.

## Notes

- `drive.yaml` is 314 lines — slightly over the ADR's "~300" target. That's the `update` op added in #105, not the split itself. Co-locating manifests under `src/services/<name>/` (the ADR's deferred alternative) would trim it; out of scope here.
- ADR-300 / ADR-100 still reference the old single `manifest.yaml` path in passing — left as historical decision records; ADR-304 supersedes that detail.


aaronsb:
## Code Review — ADR-304 manifest split

Reviewed the diff against ADR-304's migration steps and acceptance criteria. The split itself is **clean and lossless** — I verified it by round-tripping: parse the pre-split `manifest.yaml`, parse each new per-service file, reassemble into `{services: {...}}`, and compare parse-trees (order-insensitive) — they match exactly, 80 operations across 7 services. The loader rewrite and the `build`/`mcpb`/`manifest-lint` Makefile changes all hold up. One Makefile target is broken; details below.

### Blocking

**1. `make manifest-diff` doesn't run — two shell/Make escaping bugs in the new recipe** (`Makefile`, ~lines 52–64)

I ran the target (stubbing `manifest-discover` since it needs the `gws` binary) and it fails immediately:

```
[eval]:1
const body = ... .replace(/^(?!\/gm, '    '); ...
                          ^
SyntaxError: Invalid regular expression flags
```

Two separate problems:

- **Reassembly block**: `replace(/^(?!\$)/gm, '    ')` — in a Makefile recipe, `$)` is parsed as a reference to the (empty) variable named `)`, so Make expands `\$)` → `\`, and node receives `replace(/^(?!\/gm, ...)` → invalid regex. To pass a literal `$` to the recipe you need `$$`. The intent (skip empty lines so they don't get 4 spaces prepended) is correct — it just needs `replace(/^(?!$$)/gm, '    ')` so node sees `(?!$)`.
- **The `diff` line**: `diff ... "$$(node -p 'require(\"os\").tmpdir()')/..."` — inside the single-quoted shell string, `\"` stays literal, so node gets `require(\"os\")` → `SyntaxError: Expected unicode escape`. Double quotes don't need escaping inside single quotes: `node -p 'require("os").tmpdir()'`.

With both fixes applied I re-ran it: the target works, writes `$TMPDIR/gws-curated-manifest-reassembled.yaml`, and that file parses as valid YAML and is content-equivalent (order-insensitive) to the old `manifest.yaml`. Worth a `make manifest-diff` smoke check in CI or the PR checklist so this doesn't regress silently — it's a dev diagnostic with no test coverage, and this PR is the thing that rewrote it.

### Non-blocking — nits

**2. Leftover `# === <Service> ===` banner comments** — `sheets.yaml`, `docs.yaml`, `tasks.yaml`, and `meet.yaml` each open with the section-divider banner that used to separate blocks inside the monolith (e.g. `sheets.yaml` starts with a `# Sheets` banner). `gmail.yaml`/`calendar.yaml`/`drive.yaml` are clean because those blocks had no preceding banner. Harmless, but a `# Sheets` header at the top of `sheets.yaml` is redundant noise — would tidy up to drop them so all seven files start uniformly with `tool_name:`.

**3. Redundant copy in the `mcpb` target** — `cp -r build/*` already brings `build/factory/manifest/` (including `README.md`) into `mcpb/server/`, since `build` does `cp -r src/factory/manifest build/factory/manifest`. The follow-up `cp src/factory/manifest/*.yaml mcpb/server/factory/manifest/` re-copies the same files (and would error if `build/factory/manifest/` somehow didn't exist, since it doesn't `mkdir -p` the target). It's dead code now — safe to delete the line.

**4. `tools/list` ordering changes** — `loadManifest()` sorts files alphabetically (good call — `readdirSync` order is otherwise filesystem-dependent; the ADR pseudocode omits the `.sort()`), so the assembled `services` map and the `tools/list` response now go `manage_calendar, manage_docs, manage_drive, manage_email, manage_meet, manage_sheets, manage_tasks` instead of the old `manage_email, manage_calendar, ...`. `manifest-lint` output reorders too. Cosmetic — MCP clients key tools by name — but flagging in case downstream snapshots care.

### Things I checked that are fine

- **`loadManifest()` / `resolveManifestDir()`** — correct for all three runtime contexts: npx/mcpb (`_moduleDir` → `build/factory/manifest/`), dev (`_moduleDir` → `../../src/factory/manifest`), Jest CJS (`_moduleDir` null → cwd fallbacks `src/factory/manifest`, `build/factory/manifest`). The `readdirSync(...).some(f => f.endsWith('.yaml'))` probe correctly rejects non-existent paths (ENOENT caught), files-not-dirs (ENOTDIR caught), and empty/yaml-less dirs (`.some()` false). `README.md` is correctly skipped by the `.endsWith('.yaml')` filter at load time. Empirically: with `build/` absent, the registry still loads 7 services via the cwd fallback.
- **`build` idempotence** — `rm -rf build/factory/manifest && cp -r src/factory/manifest build/factory/manifest` is idempotent across repeated builds, and the `rm -rf` is *necessary* (without it, a second `cp -r` onto an existing dir would nest `build/factory/manifest/manifest/`).
- **`manifest-lint`** — works; reports the expected 80 operations / 7 services.
- **Dedent fidelity** — spot-checked sheets/docs and the nested-param indentation; the round-trip parse-equality above is the real proof.
- **Stale path references** — none. Remaining `manifest.yaml` mentions are either `discovered-manifest.yaml` (a different artifact) or intentional historical references ("the old single-file `manifest.yaml`"). ADR-100/ADR-300 left as-is is the right call (historical records; ADR-304 supersedes).
- **`src/factory/manifest/README.md`** as the new home for the contacts-exclusion / Meet-scopes notes — reasonable; the content is faithful to the old file-header comments.

### Note on test verification

The PR claims 622 tests pass; I couldn't run the suite in this worktree (no `node_modules`), so I'm taking that on faith — but for a mechanical refactor like this, the parse-tree equality check above is stronger evidence of behavioral equivalence than the test suite would be anyway.

---
Recommendation: fix the `manifest-diff` escaping (item 1) before merge; items 2–4 are optional cleanup.

*AI-assisted review.*

aaronsb:
Addressed the review in 0e76f05:
- **Blocking fix** — `make manifest-diff` now runs. The `(?!\$)` regex was getting mangled by Make (it ate the `$`), and `node -p 'require(\"os\")…'` had literal backslash-quotes inside shell single-quotes. Replaced the regex with a split/map indent and write the reassembly to a fixed `/tmp` path (this target is already bash-only dev tooling — no `os.tmpdir()` dance needed). Verified: the reassembled `services:`-wrapped file parses and diffs against `discovered-manifest.yaml`.
- Stripped the `# === Sheets ===` / etc. banner blocks the mechanical split left atop `sheets.yaml`/`docs.yaml`/`tasks.yaml`/`meet.yaml` — round-trip parse against the pre-split manifest still matches exactly.
- Removed the dead `cp src/factory/manifest/*.yaml mcpb/server/factory/manifest/` in the `mcpb` target (`cp -r build/*` already brings the dir).

The alphabetical `tools/list` ordering (from `.sort()`) is intentional — kept.

PR #123, opened 2026-05-11
feat(email): bodyFormat: 'html' returns sanitized HTML (#91, ADR-305)

### LINKED ISSUE #91: Add `bodyFormat` parameter to `get_gmail_message_content` to expose HTML body
## Problem

I use this MCP to let an LLM read and act on my Gmail. It works well for most email, but I hit a wall with any sender that puts meaningful content only in the HTML body.

The concrete case: Paperless Post invitations. The event date and time aren't in the plain-text part — they're only in the HTML. When I ask the LLM to tell me when a party is, it reads the email via `get_gmail_message_content` and reports that no date is mentioned. The information is there; the tool just never surfaces it.

This isn't a Paperless Post quirk. Marketing email, booking confirmations, and shipping notifications routinely put structured details — dates, order numbers, tracking links, confirmation codes — exclusively in HTML. The plain-text fallback is either empty or a "view this email in your browser" stub.

## Root cause

`get_gmail_message_content` fetches with `format=FULL`, so the entire MIME tree is already in memory. The HTML body (`text/html` part) is present on every call. The tool then discards it: if a `text/plain` part exists it returns that; if not, it calls `stripHtml()` on the HTML and returns the stripped result. Either way, the HTML is gone before it reaches the caller.

This is a formatting decision in the tool layer, not a Gmail API limitation.

## Proposed solution

Add an optional `bodyFormat` parameter:

```
bodyFormat?: 'plain' | 'html' | 'raw_mime'
```

Default is `'plain'`, so existing callers see no change. The HTML body is already in memory on every call — this is a filtering change, not a new API call.

## Security considerations

Raw HTML handed to an LLM is a prompt injection surface. The `'html'` mode would apply a four-layer stack: CSS hidden element stripping (via `sanitize-html`'s `exclusiveFilter` — handles nested nodes, the most common real-world injection pattern); `sanitize-html` allowlist stripping scripts, event handlers, and non-content tags; Unicode injection character removal (Tag Block U+E0000–U+E007F, BiDi overrides, zero-width chars); and Spotlighting delimiters (`<email_content source="gmail" trust="untrusted">`). Microsoft's LLMail-Inject research found Spotlighting alone drops injection success from >50% to under 2%.

## Scope

Six files: MIME body extraction, parameter schema, sanitization utility, unit tests (each sanitization layer), integration test for a no-plain-text multipart message, and updated tool description.

I have a full implementation plan ready. Happy to open a PR if this direction looks good — just say the word.

— Kelly Lewis

aaronsb:
Closes #91. Implements [ADR-305](https://github.com/aaronsb/google-workspace-mcp/blob/main/docs/architecture/api/ADR-305-shared-html-sanitization-for-agent-facing-authoring.md) (committed in `be7160f`).

## Why

`manage_email` `read` accepted only a stripped-text body. For marketing email, invitations (Paperless Post — the issue's concrete trigger), and booking confirmations whose plain-text part is a "view this email in your browser" stub, the date / tracking code / event time was unreachable through this tool — it lives only in the `text/html` part.

ADR-305 frames the change as project-wide rather than gmail-specific: the same sanitization module the scratchpad's `html` format (ADR-302), future Docs HTML export paths, and Drive HTML readers will use as those land. Per user steering: *"we need a library for sanitization and it can be used anywhere we author with the editor."*

## What

- **`src/server/formatting/html-sanitize.ts`** — the shared sanitizer. Exposes one function, `sanitizeHtmlForAgent(html, { source })`, applying the four-layer stack:
  1. CSS-hidden subtree removal (`display:none`, `visibility:hidden`, `opacity:0`, off-screen positioning, `aria-hidden="true"`) — the actual prompt-injection vehicle in published attacks against LLM email readers.
  2. Tag/attribute allowlist via `sanitize-html` — no `script`, no `style`, no event handlers; only `http`/`https`/`mailto:` on `<a>`, `http`/`https`/`cid:` on ``.
  3. Unicode injection-char strip — Tag Block (U+E0000–U+E007F), bidi overrides, zero-width chars.
  4. Spotlighting wrapper — `<gmail_content trust="untrusted">…</gmail_content>`. Microsoft's LLMail-Inject showed this alone drops injection success >50% → <2%.
- **`extractBodyFromPayload(payload, format)`** — new `format` arg. `'plain'` (default) keeps the current behavior. `'html'` prefers the `text/html` part, sanitizes it, falls back to `text/plain` only if no HTML.
- **`formatEmailDetail(data, { bodyFormat })`** — threads the format through.
- **Gmail patch `formatDetail`** — reads `ctx.params.bodyFormat` for the `read` op.
- **Manifest `read` op** — new `bodyFormat: 'plain' | 'html'` param (default `'plain'`) + a new generic `client_only: true` flag in `ParamDef` so the formatter param doesn't leak into the gws `--params` JSON. `buildResourceArgs` and `buildHelperArgs` skip client-only params.
- **Dependency** — `sanitize-html` ^2.17 + `@types/sanitize-html` ^2.16. Transitive surface is parser/serializer only (no DOM, no JS execution).

## Coverage

- **20 sanitizer unit tests** covering each layer of the stack — wrapper, allowlist (script/style stripped, event handlers stripped, dangerous URL schemes rejected, safe schemes allowed, content tags preserved), CSS-hidden subtrees (display:none/visibility:hidden/opacity:0/off-screen/aria-hidden), Unicode injection chars (zero-width / bidi / Tag Block).
- **3 integration tests** on `formatEmailDetail` — `'html'` bypasses a stub `text/plain` part, falls back to `text/plain` when no HTML, default `'plain'` unchanged.
- **645/645 tests pass**, `manifest-lint` clean (`manage_email` still 15 operations).

## What's not in scope

- `bodyFormat: 'raw_mime'` (the third mode the issue author proposed) is deferred — handing raw unparsed MIME to the agent reintroduces every threat ADR-305 mitigates. A future ADR can decide whether to expose it behind an explicit unsafe flag.
- Smart HTML → markdown conversion (return well-structured markdown instead of sanitized HTML) — worth exploring; doesn't replace the sanitization stack, so it stacks on top in a follow-up.
- Sanitizing the `'plain'` path (apply the Unicode strip + Spotlighting wrapper to existing callers) — tempting because it would harden the default, but it changes output for every existing caller. Worth a follow-up ADR once `'html'` has settled.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -40,3 +40,4 @@ _MCP tools, schemas, progressive disclosure_
 | [ADR-302](./api/ADR-302-scratchpad-html-format-and-format-aware-validation.md) | Scratchpad HTML Format and Format-Aware Validation | Draft |
 | [ADR-303](./api/ADR-303-auto-append-next-steps-in-factory-generator.md) | Auto-append next-steps in factory generator | Accepted |
 | [ADR-304](./api/ADR-304-split-manifest-yaml-into-per-service-files.md) | Split manifest.yaml into per-service files | Accepted |
+| [ADR-305](./api/ADR-305-shared-html-sanitization-for-agent-facing-authoring.md) | Shared HTML sanitization for agent-facing authoring | Accepted |
--- a/docs/architecture/api/ADR-305-shared-html-sanitization-for-agent-facing-authoring.md
+++ b/docs/architecture/api/ADR-305-shared-html-sanitization-for-agent-facing-authoring.md
@@ -1,0 +1,90 @@
+---
+status: Accepted
+date: 2026-05-11
+deciders:
+  - aaronsb
+related:
+  - ADR-301
+  - ADR-302
+---
+
+# ADR-305: Shared HTML sanitization for agent-facing authoring
+
+## Context
+
+HTML is the lingua franca of Google Workspace content, and it shows up in several places the agent needs to read, write, or edit:
+
+- **Gmail**: message bodies are HTML. The plain-text part is often a lossy stub ("view this email in your browser") for marketing email, booking confirmations, and rich notifications — meaningful content (dates, tracking codes, order numbers) lives only in the HTML part. Issue #91 is the concrete trigger: a Paperless Post invitation whose date and time were unreachable through `manage_email` `read` because the tool stripped HTML to plain text before returning. Today `extractBodyFromPayload` prefers `text/plain` over `text/html` and falls back to a crude `stripHtml()` regex when there's no plain part.
+- **Scratchpad** (ADR-302, Draft): an `html` buffer format is planned so agents can edit Gmail message bodies, Docs HTML exports, and Drive HTML files using the same line-addressed surface that handles markdown/json/csv.
+- **Docs** export: the Docs API can return HTML preserving formatting that markdown export drops; future imports of HTML into the scratchpad need the same handling.
+- **Drive**: HTML files are common; reading them into an agent today returns the raw text via the workspace, with no sanitization.
+
+There's a common requirement underneath: **any HTML that crosses into agent-facing text is a prompt-injection surface**. Style/script blocks, event handlers, CSS-hidden content (display:none, visibility:hidden, off-screen positioning, zero opacity), tag-block Unicode characters (U+E0000–U+E007F), bidi overrides (U+202A–U+202E, U+2066–U+2069), and zero-width characters have all been used in published attacks against LLM applications that consume HTML email or scraped web content. Microsoft's LLMail-Inject research (NeurIPS 2024) found that simple Spotlighting delimiters — wrapping untrusted content in a tagged block — drop injection success from >50% to under 2% even before any sanitization, and the combination is markedly stronger than either alone.
+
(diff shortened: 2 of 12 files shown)


aaronsb:
Addressed the narrow review in 181b7df — every concrete bypass the reviewer found is now closed and covered by a regression test that asserts the smuggled text does NOT survive sanitization:

| # | Bypass | Fix |
|---|---|---|
| 1 | `display:none !important` | Terminator now tolerates `!` in addition to `;` and end-of-string |
| 2 | `top:-Npx` / `bottom:-Npx` / positive `text-indent:Npx` | Position regex covers all four axes + positive direction |
| 3 | `font-size:0pt` / `0%` / `0vh` | Unit anchor dropped — any/no unit matches |
| 4 | HTML5 boolean `hidden` attribute | Added `if (attribs.hidden !== undefined) return true` |
| 5 | `width:0` / `height:0` / `max-height:0` | New collapsed-box rule |
| 6 | `clip:rect(0,0,0,0)`, `clip-path:inset(50%)`, `transform:scale(0)`, `transform:translateX(-N)` | Explicit rules for each |
| 5 (meta) | Forgeable Spotlighting wrapper | Runtime `ALLOWED_SOURCES` check at the function boundary; throws on non-allowlisted source (including the literal `gmail><script>…<x` forgery shape) |

Unchanged: opacity:0.01 left alone (legitimate styling value); `nonTextTags` redundancy and the wrapper-convention LOW item left for a follow-up. Tests: 655 pass total, 30 in the sanitizer file (was 20, +10 regression cases for these exact bypasses).

PR #129, opened 2026-05-22
feat(gmail): add attachment and html support to reply and replyAll

JeremyYowell:
## Summary

- `reply` and `replyAll` silently dropped the `attachments` parameter — it was accepted in the tool schema but never wired into the gws CLI call
- `reply` also had no `html` flag support, unlike `send`
- Neither operation had a `draft` param

## Changes

**`src/factory/manifest.yaml`**
- Added `attachments`, `html`, and `draft` params to both `reply` and `replyAll`

**`src/services/gmail/patch.ts`**
- Promoted `reply` and `replyAll` to full custom handlers in `gmailPatch`, mirroring the existing `send` handler's attachment logic:
  - Parses comma-separated attachment filenames from the workspace
  - Validates and resolves paths via `validateAttachments`
  - Passes `--attach` flags and sets `cwd` to the workspace dir (required by the gws upload endpoint)
  - Adds `--draft` when attachments are present (same behaviour as `send`)
  - Returns a clear draft ID in the response so callers know to send from Drafts

## Behaviour

Without attachments: `reply`/`replyAll` behave exactly as before — sends immediately.  
With attachments: creates a Gmail draft (consistent with `send`) and returns the draft ID.

## How to reproduce the bug

```
# File exists in workspace, replyAll silently sends without it
manage_email replyAll messageId=<id> body="see attached" attachments="report.pdf"
# → email arrives with no attachment, no error
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/factory/manifest.yaml
+++ b/src/factory/manifest.yaml
@@ -103,6 +103,15 @@ services:
             type: string
             description: "Reply body text"
             required: true
+          attachments:
+            type: string
+            description: "Workspace filenames to attach, comma-separated (files must exist in workspace via manage_workspace). Creates draft when present."
+          html:
+            type: boolean
+            description: "Treat body as HTML content (default: plain text)"
+          draft:
+            type: boolean
+            description: "Save as draft instead of sending (default: false, forced true when attachments present)"
         cli_args:
           messageId: "--message-id"
           body: "--body"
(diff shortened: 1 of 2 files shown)


aaronsb:
Thanks @JeremyYowell — good catch and a correct fix.

This branch predates the ADR-304 manifest split (`f4963c7`), so it conflicts structurally. I've cherry-picked your commit onto current `main` with your authorship intact, re-targeted the manifest hunks to `src/factory/manifest/gmail.yaml`, and added test coverage for the attachment path: #138.

Closing this in favour of that. The fix is yours.

Issue #155, opened 2026-07-30
manage_scratchpad json-mode doc import reads the first tab only


aaronsb:
`manage_scratchpad` import with `mode: 'json'` calls `documents.get` without
`includeTabsContent`, so a multi-tab document imports as its FIRST TAB ONLY —
the same defect as #152, on a different surface. `mode: 'markdown'` is fine: it
goes through Drive's `files.export`, which returns every tab.

`src/server/scratchpad/adapters/import-doc.ts:93`

This was left out of the #152 fix deliberately, because the obvious change breaks
something. Measured live: with `includeTabsContent=true`, Google **removes** `body`
from the response (along with `headers`, `lists`, `inlineObjects`, `namedStyles`)
and moves content to `tabs[].documentTab.body`.

JSON-mode import is live-bound for round-trip editing (#79), and `docs-sync`
translates mutations against paths like `$.body.content[0].paragraph.elements[0].textRun.content`.
Flip the flag and every one of those paths refers to a field that is no longer
there — so imports would gain their missing tabs and lose the ability to write
anything back.

So this needs a decision, not a one-line change:

1. **Tab-aware sync paths** — teach `docs-sync` to address
   `$.tabs[N].documentTab.body.…` and carry a tab id into the `batchUpdate`
   request. Correct, and the most work: Docs `batchUpdate` requests take a
   `tabId`/`tabsCriteria` for exactly this.
2. **Import all tabs, bind only one** — full fidelity on read, writes scoped to a
   chosen tab. Needs a way to say which.
3. **Detect and refuse** — keep the flagless call, but check tab count and tell the
   caller plainly that JSON mode covers the first tab, pointing at markdown mode
   for the whole document. Smallest honest fix; no silent truncation.

(3) is the floor — whatever else happens, this should stop presenting one tab as
the document. Related: #152, #79.

aaronsb:
Second call site with the same blind spot: `reloadDocsBuffer` in `src/server/scratchpad/handler.ts:376` also calls `documents.get` without the flag. It re-fetches the doc to refresh the live-bound buffer after a push, so it is subject to exactly the same constraint — it feeds the JSON that `docs-sync` addresses by `$.body.content[…]` paths, and the flag removes `body`.

Whatever option is chosen, both sites move together, or a scratchpad's buffer and its reload disagree about the document's shape.

aaronsb:
Released in **v4.2.0**, taking option 1 — tab-aware sync paths.

[...]

Issue #79, opened 2026-03-27
feat(scratchpad): live Docs batchUpdate sync for JSON mode


aaronsb:
## Context

ADR-301 scratchpad buffer supports JSON mode for Google Docs — import the native Docs API JSON, edit with `json_set`/`json_get`, and ideally push changes back live. Currently implemented for Sheets (via `values.update`) but Docs sync is deferred because the Docs API requires `batchUpdate` with discrete operations rather than accepting full JSON replacement.

The plumbing exists: `syncIfBound()` in `handler.ts` detects the live binding and has the hook point. The local mutation is applied to the buffer. What's missing is translating the mutation to a `batchUpdate` request.

## Proposed approach

**One edit per sync cycle with reload:**

1. `json_set` applies the mutation locally
2. `syncIfBound()` translates to `batchUpdate` operation(s)
3. API call fires
4. Full doc JSON reloaded into buffer (single source of truth)

**Two paths based on change type:**

| Change type | Example | batchUpdate translation |
|---|---|---|
| Text content | `$.body.content[2].paragraph.elements[0].textRun.content` | `deleteContentRange(start, end)` + `insertText(start, newText)` using `startIndex`/`endIndex` from the JSON structure |
| Paragraph style | `$.body.content[2].paragraph.paragraphStyle.namedStyleType` | `updateParagraphStyle` with field mask |
| Structural (add/remove elements) | Adding a new paragraph, removing a table | **Reject** with guidance: "Structural edits require markdown mode + doc_create" |

The error boundary is the `batchUpdate` API response. We don't predict failure — we try, and if the API rejects, the buffer still has the local change and the agent gets actionable guidance.

## Implementation notes

- `syncIfBound()` in `src/server/scratchpad/handler.ts` is the insertion point
- Need a `diffToRequests(oldJson, newJson)` function that compares before/after and emits batchUpdate requests
- Text content changes are the 90% case and the simplest to translate
- `startIndex`/`endIndex` are already present in the Docs JSON on every element
- Reload after push ensures the buffer always reflects actual doc state

## References

- ADR-301: Scratchpad Buffer — Service-Agnostic Content Authoring
- `src/server/scratchpad/handler.ts` lines 314-326 (current placeholder)
- Google Docs API [batchUpdate reference](https://developers.google.com/docs/api/reference/rest/v1/documents/batchUpdate)

PR #156, opened 2026-07-30
fix(docs): read every tab, not just the first

### LINKED ISSUE #152: [BUG] manage_docs get only returns the first tab of a multi-tab Google Doc
`manage_docs` with operation `get` returns the content of the first tab only. There's no error and nothing in the response to indicate the document has other tabs, so the partial content reads like the whole file.

**Steps To Reproduce**

1. Create a Google Doc with several tabs and put different text in each one.
2. Call `manage_docs` with `operation: get`, the document ID, and an authenticated account email.
3. Compare what comes back against the document. Only the first tab is there.

**Expected Behavior**

Either the actual tab that was linked, or the whole document including all tabs. Otherwise, a note in the response saying other tabs exist and are not included.

**Actual Behavior**

The first tab only, presented as the complete document. In my case the document holds multiple meeting transcripts, one per tab. `get` returned 142 lines. The full document is 780. 

Importing the same document into a scratchpad does return every tab, and reading it back by line range works well. So the data is reachable, just not through `get`.

**Environment**

- OS: macOS 26.5.2 (25F84)
- Node.js Version: bundled with the .mcpb desktop extension, not installed separately
- NPM Version: as above
- Google Workspace MCP Version: 4.0.1

**Additional Context**

I think the Docs API returns the first tab only unless `includeTabsContent=true` is passed on `documents.get`. Could `get` either set that, or flag in its response that more tabs exist?

Installed as the .mcpb desktop extension in Claude Desktop, two accounts authenticated.

aaronsb:
Closes #152.

`manage_docs get` returned the first tab of a multi-tab document and presented it
as the whole file. No error, no hint that other tabs existed — @Laralike's
document was 780 lines and came back as 142. Their diagnosis was exactly right:
Google's `body` is a legacy field holding only the first tab unless
`documents.get` is called with `includeTabsContent=true`, and we never sent it.

This is the recurring defect shape in this codebase: a read that under-reports
with no error, indistinguishable from a document that really is short. Same
family as the rate-limited inbox that rendered as rows with no sender, and as
`get` itself returning metadata and no text.

## Measured live, and two assumptions corrected

Probed three real documents before trusting the mock. Both corrections matter:

**`body` is not empty with the flag — it is absent.** Along with `headers`,
`lists`, `inlineObjects` and `namedStyles`. Content moves wholesale to
`tabs[].documentTab.body`:

| | `body` chars | `tabs` |
|---|---|---|
| `includeTabsContent` unset | 10,388 | absent |
| `includeTabsContent=true` | *(field absent)* | 10,388 + 3,832 + 698 |

So reading `tabs` **first** is load-bearing, not a preference — read `body` first
and every tabbed document renders as empty. `body` survives only as the fallback
for the flagless shape, which nothing here now requests.

**Tabs are a tree.** `childTabs` nests arbitrarily, and a walk that reads only the
top level drops nested tabs — the same bug one level down. `flattenTabs` recurses
depth-first and renders a child one heading level below its parent.

## Output

Single-tab documents report one tab titled `"Tab 1"` — a Google default, not
something a user typed — so one tab renders with **no tab scaffolding at all** and
ordinary documents look exactly as they did. Multi-tab documents gain a
`**Tabs:** N` line and a heading per tab, so stacked transcripts do not read as
one continuous document and the response states the scope it covered.

Verified end-to-end against the real 3-tab document: **14,884 characters where it
previously returned 10,213 — 31% of it was missing.** The single-tab document
renders byte-identically to before, `tabs: 1` in refs.

## Deliberately not fixed here

`importDocJson` (`src/server/scratchpad/adapters/import-doc.ts:93`) makes the same
call and has the same blind spot, but it is live-bound for round-trip editing
(#79) and `docs-sync` addresses content by `$.body.content[…]` paths. Since the
flag *removes* `body`, setting it there would trade missing tabs for broken
write-back. Filed as #155 with the three options rather than fixed blind.

## Checks

`make check`: 717 tests across 41 files (+12), type-check, lint, build,
`smoke-start`, `smoke-orphan` — all green.

### DIFF
--- a/src/__tests__/factory/docs-patch.test.ts
+++ b/src/__tests__/factory/docs-patch.test.ts
@@ -12,6 +12,11 @@
  *
  *  - `create` silently discarded a title. `title` was not in the manifest at all, so
  *    the argument vanished with no error and every document was "Untitled document".
+ *
+ *  - `get` returned THE FIRST TAB of a multi-tab document and presented it as the whole
+ *    file (#152). Google's `body` is a legacy field holding only the first tab unless
+ *    documents.get is called with includeTabsContent=true; the reporter's document was
+ *    780 lines and came back as 142, with nothing in the response saying so.
  */
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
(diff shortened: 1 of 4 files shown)


aaronsb:
Review pass applied in 07eb328. Four defects, all inside the twenty lines this PR touched, and all the same mistake the PR set out to fix — a response that understates what it read without erroring:

1. **A tab with no `documentTab` rendered as `_(this tab is empty)_`.** Three such tabs produced a document that read as blank. `FlatTab.text` is now `string | null`; unread and empty render differently, and the count of unread tabs is stated above the content.
2. **A single tab reporting no content discarded a populated `body`** — `??` never reached the fallback because `''` isn't nullish, so a document with content came back as empty. `||` fixes it and makes the handler indifferent to whether Google empties `body` (per the reference) or drops it (per measurement).
3. **`characters` and `lines` measured different strings** — the latter counted rendered headings and placeholders. Four one-line tabs reported `71 characters, 15 line(s)`. Both now count document text.
4. **`**Tabs:** N` counted nested tabs**, contradicting the Docs tab strip. Now `4 (3 top-level, 1 nested)`.

The most valuable finding was a test gap: every assertion here mocks `call()`, which sits *above* the path/query/body split. If `includeTabsContent` ever loses `location: query` in a regenerated descriptor, `splitParams` routes it to the body, `buildRequest` drops the body on a GET, and #152 returns silently with this whole suite green. `requestFor`/`queryOf` already existed for exactly this; one test now runs the real `buildRequest` against the real descriptor.

Also: `get` now warns that `insertText` indices are tab-1-relative (#157); a response with no `tabs` at all is flagged rather than presented as the document; `refs.tabTitles` carries titles even where the heading is suppressed; and both scratchpad call sites document why they must *not* set the flag (#155).

Re-verified live — 14,884 characters on the 3-tab doc, line count 241 → 233 now that scaffolding isn't counted as content, single-tab output byte-identical.

Follow-ups filed: #157 (writes can't address a tab), #158 (no way to request one tab, response unbounded).

Issue #130, opened 2026-06-03
Proposal: opt-in read-only OAuth scopes (access: 'read')


defpol:
## Proposal

Add an **opt-in** `access: 'read' | 'readwrite'` option to account authentication. When set to `'read'`, it requests the read-only variant of each service scope (`gmail.readonly`, `drive.readonly`, `calendar.readonly`, …) instead of the current read/write scopes (`gmail.modify`, `drive`, `calendar`, …).

**The default stays `'readwrite'`** — fully backward-compatible. No behavior changes unless a user opts in.

## Motivation

In multi-account setups it's common to want least privilege for some accounts — e.g. a personal or work account the assistant should only ever *read*. Today every authenticated account gets full read/write scopes. An opt-in read-only mode lets users limit the blast radius of a token (and of any injected content the model reads) without giving up the multi-account convenience.

## Sketch

- A `SERVICE_SCOPE_MAP_READONLY` alongside the existing map; any service without a read-only equivalent falls back to its current scope.
- `scopesForServices(services, access = 'readwrite')` threaded through `authenticateAccount` / `reauthWithServices`, surfaced as a non-required `access` field on the `manage_accounts` tool.

I have a working branch with this plus unit tests (a default-`readwrite` invariant test and read-mode resolution tests); full suite green locally.

## Question before opening a PR

Would you welcome this? And do you prefer one PR, or split (the scope option separate from docs)? Happy to match your house style. I could also include a short docs note on credential-file permissions and prompt-injection awareness if useful, or keep that separate.

Thanks for the project!

aaronsb:
Thanks for this, and sorry for the long silence — you asked a direct question and deserved an answer a lot sooner than this.

Short version: yes, this was the right idea, and I'm building it. Not as a way of brushing past your offer — I want to explain what changed, because the design grew past the shape either of us could have proposed from the issue alone.

Your sketch was the consent half, and it's the half I kept: `scopesForServices(services, access)` with `'readwrite'` as the default, a read-only map beside the existing one, surfaced on `manage_accounts`. `scopesForServices` really is the single choke point where service names become scope URLs, so that's the minimal change and it leaves every existing account untouched.

Two things turned up while designing around it.

[...]

PR #170, opened 2026-08-17
feat(auth): refuse writes from read-only accounts at call time (ADR-202)

aaronsb:
## Description

The **enforcement half** of ADR-202. The consent half shipped in v4.3.0 and narrows the token; until now a read-only account calling a write got Google's raw 403, naming neither the account, nor the scope, nor the way out. This closes that gap, and it's what issue #130 is currently waiting on.

```
'create' needs write access to contacts. Account someone@example.com was authorized
read-only for contacts. Re-authorize with manage_accounts {operation:'scopes',
email:'someone@example.com', services:'contacts', access:'readwrite'}, or use an
account that already has it.
```

The remedy names `contacts` — the manifest service a user would re-consent — never `people`, the Google API name they never typed and cannot use. That's why the policy needs both names, and it's asserted.

**It is always on**, unlike every other policy in this layer. The others are operator opt-ins via `GWS_SAFETY_POLICY`. This one enforces a choice the *user* already made at consent time, so gating it behind a second opt-in would make the first decorative. It's a no-op for read/write accounts, which is every account by default.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Three things found by building it

**1. Write scopes are the read/write set MINUS the read-only set.** `contacts` and `meet` both carry read-only scopes *inside* their read/write set, because the service needs them at either level. Without the subtraction, a read-only contacts account holding `directory.readonly` intersects `SERVICE_SCOPE_MAP.contacts` and the write is permitted. Found by writing the test, not by reading the code.

**2. Seven of 95 operations declare no `resource`** — pure custom handlers that make several calls — and **six of them are writes**: `gmail.send/reply/replyAll/forward`, `drive.upload`, `calendar.create`. Enforcing only where the descriptor has an entry would have exempted exactly the operations most worth enforcing. Those fall back to the manifest's `type`, which an existing test already pins against `httpMethod`.

**3. Measured across every action-typed operation:** exactly one is satisfiable by a read-only token — `drive.export`, a GET the manifest labels an action. Google accepts `drive.readonly` for `files.export`, so allowing it is **correct**. A hand-maintained list of write operations would have refused a read the token genuinely permits. ADR-103's argument arriving on its own.

## Fail-open, deliberately

No credential, unreadable credential, no scopes, no manifest info, unknown method — all allow. A safety check that blocks on its own uncertainty is an outage, and every one of these reaches Google, which refuses it independently if it should be refused.

## Interface changes, both additive

`SafetyPolicy.evaluate` may now be async and receives the operation's manifest info as an optional 4th argument. A narrower implementation still satisfies the type, so `draft-only-email`, `no-delete`, `read-only` and `audit` are untouched.

`generateHandler` takes an optional `serviceName` — the manifest key, which is the only way to reach the scope maps (`contacts`, not `people`).

## Testing Performed

- [x] 916 tests pass (was 897). 18 new in `account-access.test.ts`, plus a wiring assertion in `server.test.ts`.
- [x] Scope strings in tests are **imported from the maps**, not written out, so a test can't pass by agreeing with a URL the source has since changed.
- [x] `make check` green — typecheck, lint, gates, node-floor, build, smoke.

**Each guard proven by breaking it and watching the suite go red, then restored:**

| Break | Test that caught it |
|---|---|
| Drop the read-only subtraction in `writeScopesFor` | 2 failed — `excludes read-only scopes…`, `never grants a write scope…` |
| Drop the resource-less fallback | 1 failed — `blocks a write from a read-only account` |
| Fail *closed* on an unknown descriptor method | 1 failed — `allows when the resource is not in the descriptor` |
| Remove the always-on wiring from `server.ts` | 1 failed — `activates account-access with no GWS_SAFETY_POLICY set` |

That last one matters: without it, deleting one line in `server.ts` leaves the policy's own tests green while the server enforces nothing.

**Not yet exercised against live Google.** The policy is unit-tested with a mocked credential store; no real read-only account has hit it. That's the next task, using `aaron@bockelie.com` — and it's the honest caveat I posted on #130.

## Security Considerations

- [x] Enforcement derives from `descriptor.json` — Google's own per-method scope list — not a hand-kept table that drifts.
- [x] Fail-open is deliberate and enumerated above; none of the fail-open paths grant access Google wouldn't independently allow.
- [x] Also closes a pre-existing gap: an account that never authorized a service now gets a sentence saying so, instead of an opaque 403.
- [x] The policy reads the credential file per intercepted operation. Uncached, as ADR-202 anticipated.

## Additional Notes

ADR-202 gets a dated Update section recording what building it settled — the always-on decision, the resource-less fallback, the scope subtraction, and the `drive.export` measurement — rather than editing its original reasoning.

Still open, unchanged by this PR: `confirmWriteAccess: true` can be sent on the first call, skipping the warning, since nothing server-side remembers having warned. Worth noting the gate is currently unreachable from real input anyway — every service in the map now has a read-only scope, `meet` included via `meetings.space.readonly`. It guards the first service added without one.

### DIFF
--- a/docs/architecture/auth/ADR-202-per-account-access-level-enforced-at-call-time.md
+++ b/docs/architecture/auth/ADR-202-per-account-access-level-enforced-at-call-time.md
@@ -108,7 +108,15 @@ green while the server stopped enforcing anything.
 several calls — and six of the seven are writes (`gmail.send`, `reply`, `replyAll`,
 `forward`, `drive.upload`, `calendar.create`). Enforcing only where the descriptor has an
 entry would exempt exactly the operations most worth enforcing, so those fall back to the
-manifest's `type`, which the existing test already pins against `httpMethod`.
+manifest's `type`.
+
+That fallback was initially justified by saying the existing `type`-vs-`httpMethod` test
+already pins `type`. **It does not** — that test skips any operation with no descriptor
+method, which is these seven exactly, so `type` was unpinned precisely where it had become
+the sole enforcement input. Retyping `gmail.send` to `list` would have disabled
+enforcement for send, reply, replyAll and forward with the whole suite green. A second
+test now pins the resource-less set and their declared types by equality, so an eighth
+forces a decision instead of defaulting to unenforced.
 
 **Write scopes are the read/write set minus the read-only set,** not the read/write set.
 `contacts` and `meet` both carry read-only scopes *inside* their read/write set because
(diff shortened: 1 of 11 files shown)


aaronsb:
Review findings addressed. Summary of what changed since the review:

**Finding 2 (wiring untested) — fixed.** Two end-to-end tests now go through the real handler. Worth noting: my first attempt only covered `generateHandler` and **did not catch** the `generateTools` break — I proved that by making the break and watching it stay green. A second test goes through `generateTools`, where the wiring actually lives. Both assert the remedy names `contacts` and never `people`, and that the Google client is never called.

**Finding 3 (the `type` pin was false) — fixed.** You were right that `if (!method) continue` skips exactly the seven operations `type` is the sole input for. Now pinned by equality against the resource-less set, so an eighth forces a decision rather than defaulting to unenforced. Verified by retyping `gmail.send` to `list` and watching it go red.

**Finding 4 (cross-service overlap) — documented** in ADR-202's Update.

**Minors — fixed:** blocked-ops-as-queue-successes (with the `onError: 'bail'` case tested), stderr on a swallowed credential read, the manifest ⊆ `SERVICE_SCOPE_MAP` gate, and the module docblock, which now also names what this layer does not cover.

**Finding 1 (manage_scratchpad) — not fixed here, filed as #171.** Taking your call that it can ship documented. It's recorded in ADR-202's Update and the safety.ts docblock so neither reads as exhaustive. The issue carries both halves — the lost explanation for read-only accounts, and the sharper one you identified where `draft-only-email` is bypassed and **Google accepts**, since that has no backstop. Also carried over your `handleSend` switch suggestion, the target→resource table, the sync write-back paths outside that switch, and the point about verifying each resource resolves in the descriptor so it doesn't fail open silently.

Also confirming your verification work landed in the record: the independent re-derivation of the 95/7/6 counts, `drive.export`, `calendar.freebusy`, and the queue-dispatch check.

923 tests, up from 916.

PR #187, opened 2026-08-19
chore: security audit — patch vulnerable deps, modernize dev toolchain, pin typescript

yabdi:
## Summary

Brings the dependency tree to **`npm audit`: 0 vulnerabilities** (was 17: 1 low, 2 moderate, 14 high) and makes the build deterministic. Everything was verified with the repo's own gate: **`make check` passes end-to-end** (typecheck, lint, gate checks, 975 tests / 48 files, build + postbuild integrity, smoke, smoke-orphan) on Node 22.23.2.

## Changes

**1. Patch production-reachable vulnerabilities (`npm audit fix`, all within existing semver ranges — lockfile-only)**

- via `@modelcontextprotocol/sdk`: `fast-uri` 3.1.3→3.1.5 (host-confusion), `ip-address` 10.2.0→10.5.0 (SSRF/trust-boundary), `hono` 4.12.29→4.13.3, `body-parser` 2.2.2→2.3.0 (DoS when size limit unset), `@hono/node-server` 1.19.14→1.19.17
- via `sanitize-html`: `postcss` (sourceMappingURL path traversal), `nanoid` 3.3.12→3.3.18 (generator loop)
- also patched in-range: `flatted` 3.4.1→3.4.4, `js-yaml`, `brace-expansion`

These ship in the published package (`build/` + dependencies), so every consumer is affected.

**2. Dev toolchain: `@typescript-eslint` 6.x → 7.x**

Clears the remaining audit findings (minimatch / flatted ReDoS via the eslint chain) **without** `npm audit fix --force` (which would jump to v8). 7.x is compatible with the pinned `eslint ^8.56.0` and the existing `.eslintrc.json` (legacy format) — lint passes unchanged.

**3. Pin `typescript@^5.9.3` in devDependencies (was undeclared/transitive)**

The repo's `tsc` was only ever a transitive (hoisted 5.9.3) — `package.json` never declared its compiler. Any unconstrained re-resolution (e.g. a dependency bump) can hoist the **TypeScript 7 native preview (7.0.2)**, whose `tsc` breaks type-checking on this codebase (`TS2591: Cannot find name 'node:fs/promises'`). Pinning the compiler the build actually requires makes installs deterministic. `engines`/CI are unaffected.

**4. `src/version.ts` regenerated** by `npm run version-stamp` (comment-only sync with the current script).

## Verification

- `npm audit`: **0 vulnerabilities** (baseline was 17)
- `make check`: full gate green — typecheck, lint, check-gates, check-node-floor, check-write-ops, **975 tests / 48 files**, build + postbuild integrity check, smoke (13 tools), smoke-orphan
- No behavioral/source changes: `src/` is untouched (only `src/version.ts` comment line from the stamp)


aaronsb:
Verified and merging — thank you.

I re-measured this independently rather than taking the numbers on faith, and every claim in the description held up exactly:

- \`npm audit --package-lock-only\` on \`main\`: **17 vulnerabilities (1 low, 2 moderate, 14 high)** — your stated baseline, to the number
- Same measurement on this branch: **0**
- CI on Node 24: all 7 jobs green, **975 tests / 48 files**

What I want to call out specifically is that you did this **lockfile-only, within existing semver ranges**, and resisted `npm audit fix --force`. The five-line `package.json` diff is the correct shape for that, and the high-severity postcss path-traversals reaching production through `sanitize-html → postcss → nanoid` are genuinely fixed by the transitive bump without touching a declared range. Choosing `@typescript-eslint` 7.x over the 8.x that `--force` would have jumped to, specifically to stay compatible with the pinned `eslint ^8.56.0` and the legacy `.eslintrc.json`, is the kind of restraint that makes a security PR safe to merge quickly.

The `typescript@^5.9.3` pin is a real catch and I had not noticed the exposure. The compiler was genuinely undeclared and resolving transitively, and `latest` on the registry is now **7.0.2** — so the TS 7 native-preview hoist you describe is live today, not hypothetical.

Merging as-is. A follow-up PR from me will carry fixes for a few things I found while reviewing the other three — none of them in this one.

PR #188, opened 2026-08-19
feat(calendar): support all-day events on create and update

yabdi:
## Summary

`manage_calendar create`/`update` always sent `start`/`end` as `dateTime`, which made it impossible to create all-day events. This PR adds an `allDay: true` flag that sends Calendar API `date` fields instead, plus an all-day-aware display fix for `list`/`get`.

## Behavior

- **`allDay: true` on create/update** — `start`/`end` become dates (`YYYY-MM-DD`; an ISO datetime's date part is accepted). The Calendar API's all-day `end.date` is **exclusive** (one day after the last day), which is a classic gotcha, so the MCP treats the caller's `end` as the **inclusive last day** and computes the exclusive date: `end == start` (or a missing `end`) means a one-day event; `start=2026-08-20, end=2026-08-22` becomes `end.date=2026-08-23`.
- **`end` optional for all-day on create** — a date-only request like `{summary, start: '2026-08-20', allDay: true}` works; timed events still require `end` (validated in the handler).
- **Timed ↔ all-day conversion on update** — the field shape (`date` vs `dateTime`) is what Google keys on, so `allDay: true` with new dates converts an event to all-day, and omitting it (or `false`) with datetimes converts back.
- **Display** — `list`/`get` previously rendered an all-day event's bare `date` as a timezone-shifted local time ("Aug 19 20:00"). They now render `Thu, Aug 20 – Sat, Aug 22 (all day)` (inclusive range, `end - 1` from the API's exclusive date, timezone-proof).
- **Scratchpad** — the `calendar_event` send adapter honors the same `allDay` flag.

## Notes

- The Meet-link idempotency key now includes `allDay` so a timed and all-day event with otherwise-identical fields don't collide.
- Date validation is strict (a real calendar check, not `Date.parse`, which rolls `2026-02-30` over to March 2).

## Testing

- `make check` passes (typecheck, lint, gates, 1005 unit tests including new ones for date conversion, validation, and rendering, build, smoke).
- Verified live against the Calendar API: created and read back a multi-day all-day event (`start.date`/`end.date` fields confirmed; rendered as `Thu, Aug 20 – Sat, Aug 22 (all day)`), then deleted it.

### DIFF
--- a/src/__tests__/factory/calendar-patch.test.ts
+++ b/src/__tests__/factory/calendar-patch.test.ts
@@ -464,6 +464,121 @@ describe('calendarPatch', () => {
         { email: 'c@d.com' },
       ]);
     });
+
+    describe('all-day events (allDay: true)', () => {
+      it('sends date fields, with an EXCLUSIVE end computed from the inclusive last day', async () => {
+        // The Calendar API's all-day end date is one day AFTER the last day of
+        // the event. The caller thinks in inclusive days, so '2026-07-12' as the
+        // last day must become end.date '2026-07-13'.
+        mockCall.mockResolvedValue(calendarInsertResponse);
+        await calendarPatch.customHandlers!.create(
+          { summary: 'Vacation', start: '2026-07-12', end: '2026-07-12', allDay: true },
+          'user@test.com',
+        );
+
+        const request = await requestFor('calendar', 'events.insert', mockCall.mock.calls[0][2]);
+        expect(request.body!.start).toEqual({ date: '2026-07-12' });
+        expect(request.body!.end).toEqual({ date: '2026-07-13' });
+        expect(request.body!.dateTime).toBeUndefined();
+      });
+
+      it('extends a multi-day inclusive range by one day for the API', async () => {
+        mockCall.mockResolvedValue(calendarInsertResponse);
+        await calendarPatch.customHandlers!.create(
+          { summary: 'Conference', start: '2026-07-12', end: '2026-07-14', allDay: true },
+          'user@test.com',
+        );
+
+        const request = await requestFor('calendar', 'events.insert', mockCall.mock.calls[0][2]);
+        expect(request.body!.start).toEqual({ date: '2026-07-12' });
+        expect(request.body!.end).toEqual({ date: '2026-07-15' });
+      });
+
+      it('defaults a missing end to the start date (a one-day event)', async () => {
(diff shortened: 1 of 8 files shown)


aaronsb:
Merging — thank you. I exercised this against the live Calendar API, not just CI, and the create path is correct in a way that is genuinely hard to get right.

Verified end to end: I created a multi-day all-day event passing the **inclusive** end `2026-08-27`, confirmed Google stored the **exclusive** `2026-08-28`, and confirmed read-back rendered `Thu, Aug 27 – ...` again. The full inclusive → exclusive → inclusive round trip is right, and the weekday labels check out against `date -u`. The single-day case with `end` omitted renders `Wed, Aug 26 (all day)` with no spurious range.

Two things I want to credit directly:

- **UTC-based arithmetic throughout.** `exclusiveEndDate` parsing against `T00:00:00Z` and `formatDateLabel` building from `getUTCDay()` is what keeps the display from shifting a day under the viewer offset. That is the bug most all-day implementations ship with.
- **The strict validator.** Rejecting `2026-02-30` via UTC round-trip instead of trusting `Date.parse`, which silently rolls it to March 2, is a real edge case handled properly.

One thing did not survive live testing, and I want to be precise since it is not what a reading of the diff suggests. The **timed ↔ all-day conversion on update fails in both directions** with `400 Invalid start time` — and it fails even when both `start` and `end` are supplied, so this is not a missing-guard case:

```
update { eventId, start: "2026-08-26", allDay: true }                     -> 400
update { eventId, start: "2026-08-26", end: "2026-08-26", allDay: true }  -> 400
update { eventId, start: "...T09:00:00-06:00", end: "...T10:00:00-06:00" } -> 400
```

The cause is that `events.patch` **merges** the nested `start`/`end` objects rather than replacing them, so sending `{ date }` onto an event holding `{ dateTime }` leaves both fields set, which Google rejects. The fix is to send the counterpart as an explicit null (`{ date: "...", dateTime: null }`) so the merge clears the old shape.

Worth noting why CI could not catch this: the three update tests assert the **request body that gets built**, never a Google response — so a body the API rejects passes cleanly. That is a blind spot in the harness rather than something you missed, and your own testing notes are explicit that live verification covered create and read-back.

Merging as-is; the null-counterpart fix and the three updated assertions will come in a follow-up PR from me, credited back to this one.

PR #189, opened 2026-08-21
feat(drive): createFolder, listFolder, tree, trash, setRole

yabdi:
## What

Adds folder + recoverable-delete + silent-role-change operations to `manage_drive`, which currently cannot create folders at all.

- **`createFolder`** — create a folder (optionally nested via `parentFolderId`). `files.create` with the folder MIME type; `parents` sent as an array in the body.
- **`listFolder`** / **`tree`** — list a folder's non-trashed children, and print a recursive tree with a file count (`files.list`).
- **`trash`** — recoverable removal (`files.update` with `trashed: true`), the safe alternative to the permanent `delete`.
- **`setRole`** — change an existing collaborator's role by email (resolved to the permission id via `permissions.list` → `permissions.update`). Sends no notification email.

`files.create` and `permissions.update` are newly exposed in `docs/api-surface.md` (drive 14 → 16 of 64).

## Why

`manage_drive` can only ever produce a flat pile of files, or a zip the recipient has to unpack. Folder creation is the biggest gap; `trash` gives a recoverable alternative to the permanent `delete`; `setRole` gives the "upgrade a collaborator quietly" path (no notification email).

## Notes

- `tree` is classified as a read operation in the safety policy (`isReadOperation`).
- Tests cover request-body vs query placement against the real descriptor.


aaronsb:
Merging — thank you. I exercised every new operation against live Drive and they all work: `createFolder` standalone and nested via `parentFolderId`, `listFolder`, `tree`, and `trash`.

What I want to credit specifically:

- **The safety-policy touch is exactly one word.** Adding `tree` to the `isReadOperation` allowlist and nothing else is the right change, and `listFolder` already being covered by the `list` prefix means you did not over-reach. Touching a security-sensitive file minimally and saying so in the PR description is how that should be done.
- **`setRole` requires the permission to already exist** and gives an actionable error pointing at `share` instead. That guard is easy to skip.
- **The comments explain the non-obvious API behavior** rather than restating the code — why `parents` must be an array in the body, why `permissions.update` addresses by id rather than email. That is what makes this reviewable.
- `tree` correctly excludes subfolders from the file count, which matched the description exactly when I ran it.

Four things going into a follow-up PR from me. The first is the one I would have asked you to change before merge; I am carrying it myself to get this in quickly rather than round-tripping.

1. **`setRole` silently demotes when `role` is omitted.** The handler falls back to `reader` and the manifest declares `default: reader`, so calling it on an existing `writer` without a role downgrades them — and by this operation design, with no notification email. For an operation whose purpose is changing a role there is no safe default; it should throw the way the missing-permission branch already does. Related: `generateSchema` collects params first-wins across operations, so the `role` enum you declared on `setRole` is discarded in favor of the one `share` declares earlier. Your param descriptions show you already understood the shared-param model.

2. **`listChildren` ignores `nextPageToken`.** With `pageSize: 1000`, a folder with more children truncates silently, and both `listFolder` count and `tree` **TOTAL FILES** then report a confident wrong number. The house pattern surfaces the token and tells the model how to continue — see `meet/patch.ts:406` and `contacts/patch.ts:155`.

3. **`tree` has no depth limit or cycle guard** — one serial API call per directory with no `maxDepth` bound.

4. **Unescaped interpolation in the Drive query**: `q: "'${folderId}' in parents"`. Impact is bounded since the query only returns files the account already reads, so this is malformed-query territory rather than escalation, but `q` supports backslash escaping.

None of these blocked the merge. Nice work.

PR #190, opened 2026-08-21
feat(gmail): archive operation — save email headers + body to a workspace markdown file

yabdi:
## What

Adds an `archive` operation to `manage_email`: save one message's headers and plain-text body to a markdown file in the workspace.

- Fetch the message (`users.messages.get`, full format), extract Subject/Date/From/To/Cc and the body (prefer `text/plain`, fall back to a stripped `text/html`).
- Write `<date>_<subject>.md` to the workspace by default, or an explicit `outputPath`.
- List attachments in the file (not downloaded — `getAttachment` already does that).

## Why

A LOCAL archive — the message is read and a file is written; Gmail's INBOX label is never touched (that is `modify` with `removeLabelIds: INBOX`). This fills the "save this email as a file I can read/relay" gap that `read` (renders inline only) and `getAttachment` (attachments only) don't cover.

## Notes

- Declared `type: action` (writes a local file) over the GET `users.messages.get`, so it's added to `TYPE_EXCEPTIONS` in the generator test alongside `drive.export`.
- Unit tests cover the fetch params, header/body extraction, attachment listing, and filename slugging.


aaronsb:
Merging — thank you. I ran this against a live message and it works cleanly: headers, body, and a readable slug (`2026-08-21_foray-consulting-a-security-advisory-on-nanoid-aff.md`). Fittingly, the message I happened to test it on was the nanoid advisory that #187 patches.

The part I checked hardest was path handling, since this is the only one of your four PRs that writes to disk. What I want to credit:

- **You reused `resolveWorkspacePath` and `verifyPathSafety` instead of rolling your own.** That is the correct instinct and not the common one. `outputPath` reaching those helpers raw is fine precisely because they sanitize, resolve, and enforce workspace containment.
- **`slugify` stripping `[^\w\s-]` closes the subject-line vector.** An attacker-controlled subject of `../../etc/passwd` lands as `etcpasswd`. Since the subject comes from an untrusted sender, that mattered.
- **You anticipated the naming collision and handled it in the description the model actually reads** — "a LOCAL archive, not Gmail's 'remove from inbox'". That is the right place to disambiguate.
- Declaring `type: action` over a GET resource, and adding it to `TYPE_EXCEPTIONS` with a note, is the honest classification.

Two notes, neither blocking and neither really yours:

1. **Naming.** `modify`'s own docs say "INBOX to archive", so `archive` now means opposite things in two operations on the same tool. Your description mitigates it, but a model selecting on operation name alone could still pick wrong. Something like `saveToFile` or `export` would remove the ambiguity — your call, and I am not changing it unilaterally.

2. **Pre-existing, not introduced here:** `verifyPathSafety` early-returns on `ENOENT`, so on the write path — where the file never exists yet — the symlink check is effectively a no-op, and a dangling symlink planted inside the workspace would be followed by `writeFile`. Your usage is correct; the gap is in `workspace.ts` and belongs on its own branch. Flagging it so it is written down somewhere.

Merging as-is. Nice work — four PRs in three days, all four green, and the security one measured honestly.

PR #191, opened 2026-08-22
fix(calendar,drive): repair what live exercise caught after the #187–#190 merge

aaronsb:
Follow-up to #187–#190, as promised in the merge comments on each. All four were green and are merged; these are defects that surfaced only when the merged code ran against the live Google APIs.

## calendar — timed ↔ all-day conversion never worked

#188 advertised conversion in both directions. Neither worked:

```
update { eventId, start: "2026-08-26", allDay: true }                      -> 400 Invalid start time
update { eventId, start: "2026-08-26", end: "2026-08-26", allDay: true }   -> 400 Invalid start time
update { eventId, start: "...T09:00:00-06:00", end: "...T10:00:00-06:00" } -> 400 Invalid start time
```

`events.patch` **merges** the nested `start`/`end` objects rather than replacing them, so patching `{ date }` onto an event holding `{ dateTime }` leaves both fields set — which Google rejects. Fixed by sending the counterpart as an explicit `null`.

**Why CI could not catch it:** the three update tests assert the request body that gets built, never a Google response, so a body the API rejects passes cleanly. Updating the assertions does not close that gap — a unit test that mocks `call()` never can. The live round trip is the only real gate, and this class of defect is an argument for running one before release.

The create path in #188 is correct and stays untouched. I verified the full inclusive → exclusive → inclusive round trip live, with weekday labels checked against `date -u`.

## drive — `setRole` silently demoted

An omitted `role` fell back to `reader`, so calling `setRole` on an existing `writer` without one downgraded them — and the operation sends no notification email, so the demotion was silent. `share` may safely default to least privilege when GRANTING access; changing an existing role has no safe default. It now throws before any API call.

## drive — `listFolder` and `tree` reported confident wrong numbers

`listChildren` took the first page and dropped `nextPageToken`, so a folder above the page size truncated silently while `(n)` and `**TOTAL FILES:**` still read as totals. It now pages to exhaustion and reports truncation, so a capped count is labelled rather than asserted. This restores the house pattern from `meet/patch.ts` and `contacts/patch.ts`.

## drive — smaller hardening

- `tree` recursion is bounded by `maxDepth` (default 10) with a `seen` set. It makes one serial API call per directory and was unbounded; a Drive file may have several parents, and shortcuts can cycle.
- Single quotes interpolated into the `q` literal are escaped.

## Notes

`role` keeps **one shared declaration** across `share` and `setRole`. `generateSchema` collects params first-wins by name, and the repo has a generator test enforcing that a name is declared once — it caught my first attempt to give `setRole` its own wording, which is the guard working.

Two things found but deliberately **not** fixed here:

- `verifyPathSafety` early-returns on `ENOENT`, so on the write path the symlink check is a no-op. Pre-existing in `workspace.ts`, unrelated to #190, and belongs on its own branch.
- `manage_email.archive` collides in name with `modify`'s "INBOX to archive". #190 disambiguates in the description the model reads; renaming is a judgement call I did not want to make unilaterally on a contributor's work.

## Verification

- `make check` green — typecheck, lint, gates, build, smoke, smoke-orphan
- **1018 tests / 50 files** (4 new: pagination, `q` escaping, depth cap, cycle guard)
- `npm audit`: 0 vulnerabilities

### DIFF
--- a/src/__tests__/factory/calendar-patch.test.ts
+++ b/src/__tests__/factory/calendar-patch.test.ts
@@ -624,8 +624,10 @@ describe('calendarPatch', () => {
       );
 
       const body = (await requestFor('calendar', 'events.patch', mockCall.mock.calls[0][2])).body!;
-      expect(body.start).toEqual({ dateTime: '2026-05-01T10:00:00Z' });
-      expect(body.end).toEqual({ dateTime: '2026-05-01T11:00:00Z' });
+      // The null counterpart is what lets events.patch clear an all-day event's
+      // `date` fields; without it the merge leaves both shapes set and Google 400s.
+      expect(body.start).toEqual({ dateTime: '2026-05-01T10:00:00Z', date: null });
+      expect(body.end).toEqual({ dateTime: '2026-05-01T11:00:00Z', date: null });
     });
 
     it('maps start and end to exclusive date fields when allDay: true', async () => {
@@ -636,8 +638,8 @@ describe('calendarPatch', () => {
       );
 
       const body = (await requestFor('calendar', 'events.patch', mockCall.mock.calls[0][2])).body!;
-      expect(body.start).toEqual({ date: '2026-07-12' });
-      expect(body.end).toEqual({ date: '2026-07-15' });
+      expect(body.start).toEqual({ date: '2026-07-12', dateTime: null });
+      expect(body.end).toEqual({ date: '2026-07-15', dateTime: null });
     });
 
     it('treats an all-day end patched alone as a one-day event on that date', async () => {
@@ -651,7 +653,7 @@ describe('calendarPatch', () => {
       );
 
       const body = (await requestFor('calendar', 'events.patch', mockCall.mock.calls[0][2])).body!;
-      expect(body.end).toEqual({ date: '2026-07-21' });
+      expect(body.end).toEqual({ date: '2026-07-21', dateTime: null });
       expect(body.start).toBeUndefined();
     });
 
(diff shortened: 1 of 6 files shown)


aaronsb:
## Update: converting TO all-day now derives the missing side

The half-conversion case started as an error in both directions. Half of it turns out to be recoverable, so the behaviour is now split along where an honest answer exists.

**Converting TO all-day derives what you leave out.** A datetime projects onto a date by taking its date part, so the missing side is filled from the event on record. Deriving from the *existing* value rather than from the supplied one preserves the span:

```
event: 2026-07-18T09:00Z – 2026-07-18T10:00Z
update { eventId, end: "2026-07-20", allDay: true }
  -> start { date: "2026-07-18" }   (derived from the event)
     end   { date: "2026-07-21" }   (inclusive 07-20, exclusive +1)
```

**Converting FROM all-day still errors.** Going date to datetime means inventing a time of day that was never given, and every candidate default — 09:00, an hour long, 00:00–23:59 — silently rewrites when the event happens. That is the caller's choice, so it says so instead of guessing.

The lookup this needs still runs only when exactly one of start/end is supplied, so an ordinary update costs one API call rather than two.

Also in this push: `allDay` keeps one shared declaration across `create` and `update`. The generator test caught me diverging them — the same guard that caught the `role` divergence earlier in this PR, working twice.

Now at **1021 tests**, `make check` green.

Note for the log: `95d7f9a` is red because I pushed it before re-running the gate. `523a0f7` fixes it and is green.

